PR #2, opened 2016-02-03
Decouple batch querying from graphql-ruby integration.

dylanahsmith:
@eapache for review

## Problem

I want to test that all the computed fields on objects are being preloaded properly.  This means that all loading happens during the preload, since loading an association on an individual object will result in an N+1.

This will require exposing the ability to call the preload code without it calling the wrapped resolver, as well as the ability to execute any batch queries that are returned the preload code.  The latter is a problem in graphql-batch because the code to execute those queries was coupled with the code to integrate with graphql-ruby.

## Solution

I added a `GraphQL::Batch.execute` method that can execute a query container (single query or query group) along with any query container that is returned from a callback of that query.  I have also added a `GraphQL::Batch::QueryContainer#execute` method to make it even easier to execute queries.

Another aspect of this refactoring to remove coupling of GraphQL::Batch::ExecutionStrategy with the other code was to allow callbacks to be added on a `GraphQL::Batch::QueryContainer` after it was returned from the resolve proc.  Previously this chaining was done by passing a block to `GraphQL::Batch::Query.new` or `GraphQL::Batch::QueryGroup.new`.  Now `#then` can be used to lazily add a callback.

I also needed to handle query exceptions in GraphQL::Batch::ExecutionStrategy, so added `#rescue` and `#ensure` on GraphQL::Batch::QueryContainer to handle this in a way that seems somewhat natural for ruby.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Graphql::Batch
+# GraphQL::Batch
 
 Provides an executor for the [`graphql` gem](https://github.com/rmosolgo/graphql-ruby) which allows queries to be batched.
 
@@ -71,11 +71,11 @@ MySchema.query_execution_strategy = GraphQL::Batch::ExecutionStrategy
 
 ### Query Dependant Computed Fields
 
-If you don't want to use a query result directly, then you can pass a block which gets called after the query completes.
+If you don't want to use a query result directly, then you can use `.then` with a block to transform the query result.
 
 ```ruby
 resolve -> (obj, args, context) do
-  FindQuery.new(Product, args["id"]) do |product|
+  FindQuery.new(Product, args["id"]).then do |product|
     product.title
   end
 end
@@ -85,7 +85,7 @@ You may also need to do another query that depends on the first one to get the r
 
 ```ruby
 resolve -> (obj, args, context) do
-  FindQuery.new(Product, args["id"]) do |product|
+  FindQuery.new(Product, args["id"]).then do |product|
     FindQuery.new(Image, product.image_id)
   end
 end
@@ -95,13 +95,54 @@ If the second query doesn't depend on the other one, then you can use GraphQL::B
 
 ```ruby
 resolve -> (obj, args, context) do
-  smart_collection_query = CountQuery.new(SmartCollection, context.shop_id)
-  custom_collection_query = CountQuery.new(CustomCollection, context.shop_id)
-
-  QueryGroup.new([smart_collection_query, custom_collection_query]) do
-    smart_collection_query.result + custom_collection_query.result
+  QueryGroup.new([smart_collection_query, custom_collection_query]).then do |results|
+    results.reduce(&:+)
   end
 end
+
+### Error Handling
+
+Exceptions can be rescued by using `.rescue(error_class)` instead of `.then` which could be used for a fallback
+
+```ruby
+resolve -> (obj, args, context) do
+  CacheFetchQuery.new(Product, args["id"]).rescue(Redis::BaseConnectionError) do |err|
+    logger.warn err.message
+    FindQuery.new(Product, args["id"])
+  end
+end
+```
+
+If you want to do something after the query without affecting the result, then you can use `.ensure` which could be used for instrumentation
+
+```ruby
+resolve -> (obj, args, context) do
+  t0 = Time.now
+  FindQuery.new(Product, args["id"]).ensure do |result, error|
+    duration = Time.now - t0
+    logger.info "Product load completed in #{duration} seconds"
+  end
+end
+```
+
+## Unit Testing
+
+Batch query objects have an execute method to simplify unit testing and debugging by allowing batch queries to be executed without having to define a graphql schema.
+
+```ruby
+  def test_single_query
+    product = products(:snowboard)
+    assert_equal product.title, FindQuery.new(Product, args["id"]).then(&:title).execute
+  end
+
+  def test_batch_query
+    products = [products(:snowboard), products(:jacket)]
+    query1 = FindQuery.new(Product, products(:snowboard).id).then(&:title)
+    query2 = FindQuery.new(Product, products(:jacket).id).then(&:title)
+    group_query = QueryGroup.new([query1, query2]).execute
+    assert_equal products(:snowboard).title, query1.result
+    assert_equal products(:jacket).title, query2.result
+  end
 ```
 
 ## Development
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -1,6 +1,48 @@
 require "graphql"
+
+module GraphQL
+  module Batch
+    module_function
+
+    def execute(container)
+      unless container.is_a?(QueryContainer)
+        return container
+      end
+      if container.completed?
+        raise container.error if container.error
+        return container.result
+      end
+
+      batched_queries = Hash.new{ |hash, key| hash[key] = [] }
+      register_queries = lambda do |query_container|
+        query_container.each_query do |query|
+          batched_queries[query.group_key] << query
+        end
+      end
+      register_queries.call(container)
+
+      until batched_queries.empty?
+        queries = batched_queries.shift.last
+        begin
+          queries.first.class.execute(queries)
+        rescue => err
+          queries.each do |query|
+            query.complete(error: err) unless query.completed?
+          end
+        end
+        queries.each do |query|
+          register_queries.call(query)
+        end
+      end
+      raise container.error if container.error
+      container.result
+    end
+  end
+end
+
 require_relative "batch/version"
 require_relative "batch/query_container"
+require_relative "batch/query_callback"
 require_relative "batch/query"
 require_relative "batch/query_group"
 require_relative "batch/execution_strategy"
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -1,56 +1,61 @@
 module GraphQL::Batch
   class ExecutionStrategy < GraphQL::Query::SerialExecution
-    attr_reader :batched_queries
-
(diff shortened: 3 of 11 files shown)


rmosolgo:
wow, seems like a big win!

README.md
```diff
@@ -95,15 +95,56 @@ If the second query doesn't depend on the other one, then you can use GraphQL::B
...
+```ruby
+resolve -> (obj, args, context) do
+  CacheFetchQuery.new(Product, args["id"]).rescue(Redis::BaseConnectionError) do |err|
+    logger.warn err.message
+    FindQuery.new(Product, args["id"])
   end
 end
 ```
 
+If you want to do something after the query without affecting the result, then you can use `.ensure` which could be used for instrumentation
```
eapache: I'm assuming `ensure` is run even if an exception occurs... might want to make that explicit

lib/graphql/batch.rb
```diff
@@ -1,6 +1,48 @@
 require "graphql"
+
+module GraphQL
+  module Batch
+    module_function
+
+    def execute(container)
+      unless container.is_a?(QueryContainer)
+        return container
```
eapache: this can go on one line

eapache:
As per our discussion IRL this feels an awful lot like hand-rolled promises... is there a ruby promise library we could use to get all that functionality for free?

lib/graphql/batch.rb
```diff
@@ -1,6 +1,48 @@
...
+      batched_queries = Hash.new{ |hash, key| hash[key] = [] }
+      register_queries = lambda do |query_container|
+        query_container.each_query do |query|
+          batched_queries[query.group_key] << query
+        end
+      end
+      register_queries.call(container)
+
+      until batched_queries.empty?
+        queries = batched_queries.shift.last
```
eapache: `last` here is used to isolate the value, because `shift` on a hash gives `[key, value]` right? is there a clearer syntax for this? (`.values.shift` is clearer but presumably doesn't actually remove the value from the hash)

eapache: Ah, this should really be an `.each` except we're doing arbitrary mutations to the hash in the loop... yuck

lib/graphql/batch.rb
```diff
@@ -1,6 +1,48 @@
...
+        begin
+          queries.first.class.execute(queries)
+        rescue => err
+          queries.each do |query|
+            query.complete(error: err) unless query.completed?
+          end
+        end
+        queries.each do |query|
+          register_queries.call(query)
```
eapache: doesn't this loop forever... you're unconditionally adding back all the queries we just executed

lib/graphql/batch.rb
```diff
@@ -1,6 +1,48 @@
...
+        end
+      end
+      raise container.error if container.error
```
eapache: not obvious what would be setting this?

lib/graphql/batch.rb
```diff
@@ -1,6 +1,48 @@
...
+      unless container.is_a?(QueryContainer)
+        return container
+      end
+      if container.completed?
+        raise container.error if container.error
+        return container.result
+      end
+
+      batched_queries = Hash.new{ |hash, key| hash[key] = [] }
+      register_queries = lambda do |query_container|
```
eapache: Does it make sense to define an `Executor` class which has `register_queries` as a method and `batched_queries` as an instance var? I think it would break up the complexity a bit.

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
+
+      private
+
+      def wrap_queries(obj)
+        queries = nil
+        case obj
+        when Array
+          array = obj
+          array.each_with_index do |value, i|
+            value = wrap_queries(value)
```
eapache: the inside of these two iterators are identical and should be extractable into a method

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
+              end
+            end
+          end
+        when Hash
+          hash = obj
+          hash.each do |key, value|
+            value = wrap_queries(value)
+            if value.is_a?(QueryContainer)
+              hash[key] = nil
+              queries ||= []
```
eapache: you can just assign `queries = []` up top and check `queries.empty?` in your ternary at the bottom

dylanahsmith:
Yes, the API is a lot like promises, but we don't want a promise library that executes the block asynchronously (e.g. using event machine) or concurrently (e.g. in a separate thread).

dylanahsmith:
It looks like https://rubygems.org/gems/promise.rb might work, since it doesn't have any dependancies, so would work with the deferred semantics I need.  Using a generic promise implementation might also allow this gem to be adapted to the asynchronous or concurrent use cases.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: I know. I was just trying to avoid extra object allocations, since this gets called on every value.

README.md
dylanahsmith: Oops.  I meant to mention that.

lib/graphql/batch.rb
dylanahsmith: This just registers the new queries.

eapache: Oh, I see because `each_query` no longer unconditionally yields `self`... that was subtle

eapache: Perhaps there should be a `subqueries` method which only ever returns queries that are not `self` so we can call `register_queries.call(query.subqueries)` which is much more obvious?

Or, more deeply, now that even single queries can contain subqueries, perhaps the class structure of groups and containers needs to be rethought?

lib/graphql/batch/query_callback.rb
```diff
@@ -0,0 +1,16 @@
...
+  class QueryCallback < QueryContainer
+    def initialize(query, &block)
+      super(query)
+      @block = block
+    end
+
+    def call(result, error)
+      begin
+        complete(@block.call(result, error))
+      rescue => block_error
```
eapache: I'm wondering if this should be broader than `StandardError` or not, just in case someone tries to `.rescue(Exception)` for some reason? I know it's usually considered bad form to rescue from anything higher than `StandardError`, so just curious if that's still true for promise-y stuff like this?

(may be irrelevant if we end up using a promise library)

lib/graphql/batch/query_group.rb
```diff
@@ -1,35 +1,48 @@
...
+        if query.is_a?(QueryContainer)
+          @pending_queries.add(query)
+        else
+          @results[i] = query
+        end
+      end
+
+      if @pending_queries.empty?
+        complete(@results)
+      else
```
eapache: put a return here to unindent the large `else` block

lib/graphql/batch/query_group.rb
```diff
@@ -1,35 +1,48 @@
...
+      raise ArgumentError, "QueryGroup.new no longer takes a block, use #then instead" if block_given?
+      super(self)
+      @pending_queries = Set.new
+      @results = []
+
+      queries.each_with_index do |query, i|
+        if query.is_a?(QueryContainer)
+          @pending_queries.add(query)
+        else
+          @results[i] = query
```
eapache: Does ruby just automatically expand the array when assigning to an index past the end?

lib/graphql/batch/query_group.rb
```diff
@@ -1,35 +1,48 @@
...
+        queries.each_with_index do |query, i|
+          next unless query.is_a?(QueryContainer)
+          query.then do |result|
+            @results[i] = result
+            @pending_queries.delete(query)
+            if @pending_queries.empty?
+              complete(@results)
+            end
+          end.rescue do |err|
+            unless completed?
```
eapache: when could this be called and we're already completed? If we're completed, then all our children have returned results, and so none of them could raise the error?

lib/graphql/batch.rb
dylanahsmith: I'm planning on having a global (thread local) executor where queries will be registered so it doesn't have to be tracked on the Query object.  So far I have https://github.com/shopify/graphql-batch/compare/promise which uses promise.rb for promises, but I want to move the query registering out of Query.new.  I think I'll end up changing the API to be `CustomLoader.for(*group_args).load(key)` where `.for(*group_args)` will create and register a loader if one doesn't exist on the executor and `.load(key)` will add a key to the loader and return a Promise for that key.

lib/graphql/batch/query_group.rb
dylanahsmith: Yup, it expands it with `nil` entries when assigning past the end of an array.

```
irb(main):001:0> a = []
=> []
irb(main):002:0> a[100] = 1
=> 1
irb(main):003:0> a
=> [nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, nil, 1]
```

lib/graphql/batch/query_group.rb
dylanahsmith: right, this can be removed

PR #3, opened 2016-02-04
Use Promise.rb to provide a standard promise based API.

dylanahsmith:
@eapache for review
cc @rmosolgo
Closes #2 

This addresses the same issues as pull #2 by providing GraphQL::Batch.wait can be used to wait for and get the result of a query.  However, in this pull request I am using [promise.rb gem](https://rubygems.org/gems/promise.rb) to try to simplify the code and provide a more standard API.

I have also switched away from using `CustomQuery.new(*args)` for creating a query object to using `CustomLoader.for(*group_args).load(key)`.  This is because we now need to register the query with an executor for grouping, which I didn't want to do in `GraphQL::Batch::Query.new`.  Instead, we get a loader from the executor using the group args, or create and register the loader if it doesn't already exist.

In the future we could split the batch loading into a separate gem, so graphql-batch only provides the integration with graphql-ruby.  We could also allow a custom executor be used so that data can be loaded asynchronously or concurrently.


README.md
```diff
@@ -85,23 +78,55 @@ You may also need to do another query that depends on the first one to get the r
...
+If the second query doesn't depend on the first one, then you can use GraphQL::Batch.all, which allows each query in the group to be batched with other queries.
 
 ```ruby
 resolve -> (obj, args, context) do
-  smart_collection_query = CountQuery.new(SmartCollection, context.shop_id)
-  custom_collection_query = CountQuery.new(CustomCollection, context.shop_id)
-
-  QueryGroup.new([smart_collection_query, custom_collection_query]) do
-    smart_collection_query.result + custom_collection_query.result
+  GraphQL::Batch.all([
```
eapache: should this take a splat to avoid the extra array?

README.md
```diff
@@ -28,54 +28,47 @@ Require the library
...
+    super()
     @model = model
-    @id = id
-    super(&block)
   end
 
-  # super returns the class name
-  def group_key
-    "#{super}:#{model.name}"
+  def load(key)
```
eapache: I wouldn't bother documenting this, I imagine it will be unnecessary 99% of the time and doesn't do anything that isn't obvious with usual ruby method overrides?

README.md
```diff
@@ -28,54 +28,47 @@ Require the library
...
+Define a custom loader, which is initialized with arguments that are used for grouping and an execute method for doing the batch query. The load method can be overriden to coerce the type of the key.
 
 ```ruby
-class FindQuery < GraphQL::Batch::Query
-  attr_reader :model, :id
+class RecordLoader < GraphQL::Batch::Loader
+  attr_reader :model
 
-  def initialize(model, id, &block)
+  def initialize(model)
```
eapache: I imagine we don't want people instantiating these directly... can we give this some other name (`set_group_key` or something?) and lock down `initialize` so that calling it directly raises?

README.md
```diff
@@ -28,54 +28,47 @@ Require the library
 require 'graphql/batch'
 ```
 
-Define a GraphQL::Batch::Query derived class. Use group_key to specify which queries can be reduced into a batch query, and an execute class method which makes that batch query and passes the result to each individual query.
+Define a custom loader, which is initialized with arguments that are used for grouping and an execute method for doing the batch query. The load method can be overriden to coerce the type of the key.
 
 ```ruby
-class FindQuery < GraphQL::Batch::Query
-  attr_reader :model, :id
+class RecordLoader < GraphQL::Batch::Loader
```
eapache: Not super-keen on the name `Loader`... perhaps `Resolver` instead, with a `resolve` method instead of `execute`?

README.md
```diff
@@ -28,54 +28,47 @@ Require the library
...
 
-  def self.execute(queries)
-    model = queries.first.model
-    ids = queries.map(&:id)
-    records_by_id = model.where(id: ids).index_by(&:id)
-    queries.each do |query|
-      query.complete(records_by_id[query.id])
-    end
+  def execute
+    model.find(keys).each { |record| fulfill(record.id, record) }
```
eapache: is `keys` just a method that's available to you as a subclass? we should either document "use `keys` to access the batch of arguments" or just pass it into this method as a param

README.md
```diff
@@ -28,54 +28,47 @@ Require the library
...
+Use the batch execution strategy with your schema
 
 ```ruby
-resolve -> (obj, args, context) { FindQuery.new(Product, args["id"]) }
+MySchema = GraphQL::Schema.new(query: MyQueryType)
+MySchema.query_execution_strategy = GraphQL::Batch::ExecutionStrategy
 ```
 
-Use the batch execution strategy with your schema
+When defining your schema, using the graphql gem, return the result of calling `.load` on your loader object from the resolve proc.
```
eapache: Also need to document use of `for` and the group keys instead of instantiating a new one

README.md
```diff
@@ -85,23 +78,55 @@ You may also need to do another query that depends on the first one to get the r
...
-    smart_collection_query.result + custom_collection_query.result
+  GraphQL::Batch.all([
+    CountLoader.for(Shop, :smart_collections).load(context.shop_id),
+    CountLoader.for(Shop, :custom_collections).load(context.shop_id),
+  ]).then do |results|
+    results.reduce(&:+)
   end
 end
+
+`.then` can optionally take two lambda arguments, the first of which is equivalent to passing a block to `.then`, and the second one handles exceptions.  This can be used to provide a fallback
```
eapache: what happens if you give `then` a first argument *and* a block?

wording tweaks:
> `.then` can optionally take two lambda arguments instead of a block. The first argument takes the place of a block (being called on success) and the second is called when an exception occurs. This can be used to provide a fallback:

README.md
```diff
@@ -85,23 +78,55 @@ You may also need to do another query that depends on the first one to get the r
...
+    raise exc unless exc.is_a?(Redis::BaseConnectionError)
+    logger.warn err.message
+    RecordLoader.for(Product).load(args["id"])
+  end)
+end
+```
+
+## Unit Testing
+
+GraphQL::Batch.wait can be used to wait for a promise to be resolved and return its result. This can be useful for debugging and unit testing loaders.
```
eapache: To me `wait` implies it will wait until the batch has resolved, but will not necessarily trigger the resolution of that promise, especially when we have a separate `GraphQL::Batch.execute` which sounds like it ought to do just that... what is the difference between these two methods?

README.md
dylanahsmith: This is a contrived example, so I'm not sure if it really represents real usage.  It is only used in 3 places in Shopify, and all those places are mapping an array to provide input, so that change would actually make it less convenient.

README.md
dylanahsmith: Sure, I can leave it undocumented, since we don't need to explain how to override a method in ruby.

lib/graphql/batch.rb
```diff
@@ -1,6 +1,44 @@
...
+      obj.is_a?(Promise) ? obj : Promise.new.tap { |d| d.fulfill(obj) }
+    end
+
+    def self.all(enumerable)
+      remaining = 0
+      promise = Promise.new
+      on_fulfill = lambda do |result|
+        remaining -= 1
+        if remaining.zero?
+          promise.fulfill(enumerable.map { |obj| obj.is_a?(Promise) ? obj.value : obj })
```
eapache: does `all` need to support a mix of promise and non-promise arguments? what is the use case for passing in non-promise args here?

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
+      private
+
+      def lazy_resolve_object(obj)
+        queries = nil
+        case obj
+        when Array
+          array = obj
+          array.each_with_index do |value, i|
+            if query = lazy_key_assign(array, i, value)
+              (queries ||= []) << query
```
eapache: this is gross, do the unoptimized version unless we have benchmarks showing the extra object allocations are actually a problem

README.md
dylanahsmith: Yup, it is just an instance method that gives access to the keys passed into `#load`.  Passing it as an argument might make sense, since then it can be given a more appropriate name (e.g. `ids` in this case).

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
+      def lazy_key_assign(obj, key, value)
+        value = lazy_resolve_object(value)
+        if value.is_a?(Promise)
+          obj[key] = nil
+          query = value.then do |result|
+            obj[key] = result
           end
         end
-        result_hash
+        query
```
eapache: this is `nil` outside the `if`? so I think you can `return unless value.is_a?(Promise)` a few lines up?

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
-      attr_accessor :result_hash
-
       def get_finished_value(raw_value)
-        if raw_value.is_a?(QueryContainer)
-          raw_value.query_listener = self
-          register_queries(raw_value)
-          self
+        if raw_value.is_a?(Promise)
+          raw_value.then(->(result) { super(result) },
+                         ->(exc) { exc.is_a?(GraphQL::ExecutionError) ? super(exc) : raise(exc) })
```
eapache: exception-handling in promises makes my head hurt... the `raise` case will just re-set the error condition on the promise, while the `super` case will end up setting the exception itself in the response hash right? Why are we doing that?

lib/graphql/batch/executor.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def initialize
+      @loaders = {}
+    end
+
+    def shift
+      @loaders.shift.last
+    end
+
+    def execute(loader)
```
eapache: make `execute` and `shift` private? or maybe fold some of that together, right now `execute` could even be static which feels a bit weird

README.md
eapache: Why is the inheritor even responsible for storing the group key, shouldn't we just do that for free and pass it into the `execute` (or whatever it's called) method?

lib/graphql/batch.rb
eapache: Oh, I see the use case in `lazy_resolve_object`

lib/graphql/batch/execution_strategy.rb
dylanahsmith: I could add benchmarks, but I don't think that is a high priority right now, since a contrived benchmarks can be quite misleading.

I'm not crazy for considering performance considering that liquid and identity cache have generally been at the top of our CPU profiles in production, and removing object allocation has been the primary way of improving those code paths.

lib/graphql/batch.rb
eapache: kind of surprised `promise.rb` doesn't provide this method tbh

README.md
dylanahsmith: `GraphQL::Batch.resolve` `GraphQL::Batch.all` are based on the javascript methods [`Promise.resolve`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve) and [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all).

> The Promise.resolve(value) method returns a Promise object that is resolved with the given value. If the value is a thenable (i.e. has a then method), the returned promise will "follow" that thenable, adopting its eventual state; otherwise the returned promise will be fulfilled with the value.

There isn't a standard method name to wait for the result of a promise.

lib/graphql/batch/executor.rb
```diff
@@ -0,0 +1,38 @@
...
+      loader.execute
+    rescue => exc
+      loader.reject_all(exc)
```
eapache: this now rejects even the promises that had been fulfilled prior to the exception... is that intended or does `promise.rb` make that a no-op so we don't care?

lib/graphql/batch.rb
dylanahsmith: I'll open an upstream pull request to add Promise.all and Promise.resolve

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
+        case obj
+        when Array
+          array = obj
+          array.each_with_index do |value, i|
+            if query = lazy_key_assign(array, i, value)
+              (queries ||= []) << query
+            end
+          end
+        when Hash
+          hash = obj
```
eapache: all this does is give it a new name... doesn't feel necessary since most of the usage is now in `lazy_key_assign` anyways?

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,62 @@
...
-            value.result_hash = result_hash
+        lazy_resolve_object(super)
+      end
+
+      private
+
+      def lazy_resolve_object(obj)
+        queries = nil
+        case obj
+        when Array
```
eapache: thinking about this a little longer... when does a selection ever even result in array... aren't they *always* hashes, and arrays are just fields?

lib/graphql/batch/executor.rb
dylanahsmith: Yup, it does nothing if it isn't pending (https://github.com/lgierth/promise.rb/blob/v0.6.1/lib/promise.rb#L61-L75) because the [promise spec](https://promisesaplus.com/#point-14) says that a promise can't transition from the fulfilled or rejected state.

README.md
dylanahsmith: I thought calling it a loader was appropriate, since this library is similar to the [dataloader js library](https://github.com/facebook/dataloader).  I don't see why Loader isn't an appropriate name, since it is loading a record.

eapache: Nothing wrong with loader, I just thought resolver might be more specific because its primary purpose is to resolve (err... fulfill) the promises it collects... the fact that it usually loads data to do that isn't fundamental

README.md
dylanahsmith: Extracting the arguments from an array in execute will make the execute method less readable and delay errors from providing the wrong number of arguments.  The base Loader class already uses the grouping arguments as a key for the loader, but the derived class still needs to get what it needs to actually do the query.

dylanahsmith: I would prefer to keep the initialization in `initialize`, since that is exactly what the method is doing.  It isn't dangerous the construct the object, and could be used for unit testing or validation (e.g. check that an association exists when defining the schema).

eapache: > It isn't dangerous to construct the object

Oh OK, I kind of assumed it was

README.md
dylanahsmith: It would ignore the block if both are specified `on_fulfill ||= block`.  Sure, I can tweak it to make that more clear

lib/graphql/batch/execution_strategy.rb
dylanahsmith: No, you can have a selection on an array field.  E.g. `products(first: 50) { id }` is a selection which would resolve to an array.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: Yup, I'll do that.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: Oddly graphql-ruby expects a GraphQL::ExecutionError to be returned rather than raised (https://github.com/rmosolgo/graphql-ruby/blob/v0.10.9/lib/graphql/query/serial_execution/field_resolution.rb#L30).  @rmosolgo why doesn't graphql-ruby allow a `GraphQL::ExecutionError` to be raised from a resolve proc to specify an error?

There is support for doing exception handing in a middleware, but the middleware is used in `get_raw_value`, which has already returned, so isn't very useful for error handling in promises.

lib/graphql/batch/executor.rb
dylanahsmith: I started to split up the methods, since I plan on allowing a custom Executor to be specified.  For instance, a multi-threaded executor would want to add locking around `shift`, and would want to be able to replace `wait` so it can spawn multiple threads to execute multiple loaders in parallel.

I guess `execute` could probably move to the `Loader` class, but I don't want the custom Loader class to have to do this error handling, so I'll have to choose another appropriate name.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: Yup, I can just use `obj`

lib/graphql/batch/execution_strategy.rb
rmosolgo: No good reason, I think! It seemed simpler, but I guess it's not idiomatic Ruby. Switching to raise & rescue would be good change!

dylanahsmith:
@eapache could you re-review this.  I think all the comments have been addressed now.

README.md
```diff
@@ -28,54 +28,41 @@ Require the library
 require 'graphql/batch'
 ```
 
-Define a GraphQL::Batch::Query derived class. Use group_key to specify which queries can be reduced into a batch query, and an execute class method which makes that batch query and passes the result to each individual query.
+Define a custom loader, which is initialized with arguments that are used for grouping and an execute method for doing the batch query.
```
eapache: it seems to be called `perform` now, not `execute`

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,57 @@
...
+          obj.each do |key, value|
+            each_promise_in_entry(obj, key, value, &block)
           end
         end
-        result_hash
+      end
+
+      def each_promise_in_entry(obj, key, value, &block)
+        yield obj, key, value if value.is_a?(::Promise)
+        each_promise(value, &block)
```
eapache: there's no point in calling this if we yielded above right, should it be in an if/else?

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,57 +1,57 @@
...
-          if value.is_a?(FieldResolution)
-            value.result_hash = result_hash
+        as_promise_unless_resolved(super)
+      end
+
+      private
+
+      def as_promise_unless_resolved(result)
+        all_promises = []
+        each_promise(result) do |obj, key, promise|
```
eapache: what if `result` is a single promise, should this call `each_promise_in_entry` instead? or can that not happen?

eapache: Oh, it can't happen this is in `SelectionResolution` not `FieldResolution`, my mistake

eapache:
A few minor tweaks, but this is *so* much easier to follow now, :+1:

README.md
dylanahsmith: Right, I'll update it.  I decided the name `execute` sounded better for a query rather than a loader.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: You're right.  I changed it to use `if``else`.

PR #4, opened 2016-02-16
fix: Wait for subpromises after a list promise is resolved.

dylanahsmith:
@eapache for review

## Problem

There was a bug in https://github.com/Shopify/graphql-batch/pull/3 with waiting for promises in the result of a selection for a list.  E.g. a selection for products would resolve to a list of promises for the variants if any batch loads were needed for variant fields.

## Solution

After assigning the result of a promise for a selection to a field, I recursively called the code to turn a value into a promise to return from the `.then` block so that the outer `Promise.all` would wait for that promise.

Previously the wrapping of values with promises was handled in the selection resolution for each object.  However, since we need to handle sub-promises, we can move the promise wrapping so that it is done once (recursively) for the whole operation.  I think this also simplifies the code and makes this type of bug harder to miss.

### DIFF
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -1,12 +1,16 @@
 module GraphQL::Batch
   class ExecutionStrategy < GraphQL::Query::SerialExecution
     def execute(_, _, _)
-      as_promise_unless_resolved(super).sync
+      as_promise(super).sync
     ensure
       GraphQL::Batch::Executor.current.clear
     end
 
     private
+
+    def as_promise(result)
+      GraphQL::Batch::Promise.resolve(as_promise_unless_resolved(result))
+    end
 
     def as_promise_unless_resolved(result)
       all_promises = []
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -19,7 +19,10 @@ module GraphQL::Batch
         all_promises = []
         each_promise(result) do |obj, key, promise|
           obj[key] = nil
-          all_promises << promise.then { |value| obj[key] = value }
+          all_promises << promise.then { |value|
+            obj[key] = value
+            as_promise_unless_resolved(value)
+          }
         end
         return result if all_promises.empty?
         Promise.all(all_promises).then { result }
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -1,52 +1,43 @@
 module GraphQL::Batch
   class ExecutionStrategy < GraphQL::Query::SerialExecution
-    class OperationResolution < GraphQL::Query::SerialExecution::OperationResolution
-      def result
-        GraphQL::Batch::Promise.resolve(super).sync
-      ensure
-        GraphQL::Batch::Executor.current.clear
+    def execute(_, _, _)
+      as_promise(super).sync
+    ensure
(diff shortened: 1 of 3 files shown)


lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,49 +1,43 @@
...
-        each_promise(result) do |obj, key, promise|
-          obj[key] = nil
-          all_promises << promise.then { |value| obj[key] = value }
+    def as_promise(result)
+      all_promises = []
+      each_promise(result) do |obj, key, promise|
+        obj[key] = nil
+        all_promises << promise.then do |value|
+          obj[key] = value
+          as_promise(result)
```
eapache: this seems like it is correct but also could recurse much more than is truly necessary... I *think* it makes sense to do it in the `then` block of the `Promise.all` if `all_promises` wasn't empty?

eapache:
I think I'm missing why regular promise chaining didn't take care of this problem automatically... shouldn't any promises returned from `then` blocks be automatically chained?

lib/graphql/batch/execution_strategy.rb
dylanahsmith: Oops, this should have been `as_promise(result)`.  I guess I should also keep the method as as_promise_unless_resolved so that it doesn't bother returning a promise and depend on a check for whether the promise is pending.

dylanahsmith:
> shouldn't any promises returned from `then` blocks be automatically chained?

Yes, but it won't automatically chain an array of promises being returned from a `then` block, which is why it didn't get caught by the existing tests.

lib/graphql/batch/execution_strategy.rb
dylanahsmith: updated to address this comment.

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,49 +1,44 @@
...
+      each_promise(result) do |obj, key, promise|
+        obj[key] = nil
+        all_promises << promise.then do |value|
+          obj[key] = value
+          as_promise_unless_resolved(value)
         end
-        return result if all_promises.empty?
-        Promise.all(all_promises).then { result }
       end
+      return result if all_promises.empty?
```
eapache: how does this work, `result` isn't guaranteed to implement `sync`?

dylanahsmith: Right, I need to use `GraphQL::Batch::Promise.resolve(super).sync` again

dylanahsmith: Fixed and added a regression test.

eapache: regression test is good but I don't know why you added this line in the first place it doesn't win us much of anything?

dylanahsmith: Previously the code was unconditionally returning a promise in the `.then` block.  I'm actually surprised that didn't result in infinite recursion.  Relying on that behaviour seems fragile and confusing to read.

eapache: Oh, you don't need this line or the separate `as_promise` method if you just move the `sync` call to the bottom here, e.g. `Promise.all(...).then { ... }.sync`

eapache: Not entirely sure what the method should be called in that case though... I think it ends up cleaner in any case

eapache:
OK, thought on simplifying (based on moving the `sync`) aside, this LGTM

PR #5, opened 2016-02-17
Improve error handling

dylanahsmith:
@eapache for review

## Problem

I noticed a problem with the RecordLoader example in the README when records were missing.  We weren't fulfilling the load for ids if they are missing in the query result.

When we do come across a bug in a loader, we should also make these errors easier to debug by checking for unfulfilled promises in the loader and in the executor to provide a helpful error message.

I also noticed that we aren't actually handling a GraphQL::ExecutionError properly, since it wasn't behaving as if the error was raised in a field's resolve proc.  The error was causing the whole request to fail, without showing other fields, instead of being treated as a field error.

## Solution

* I added `GraphQL::Batch::Loader#fulfilled?` to make it easier to check for unfulfilled promises in `#perform`
* GraphQL::Batch::Loader#resolve now raises GraphQL::Batch::BrokenPromiseError if a promise is still pending after `#perform`
* GraphQL::Batch::Executor#wait now raises GraphQL::Batch::BrokenPromiseError if the promise is still pending after all loads have been performed
* I added an on_reject handler in the get_finished_value override in the execution strategy

### DIFF
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -49,7 +49,12 @@ module GraphQL::Batch
     class FieldResolution < GraphQL::Query::SerialExecution::FieldResolution
       def get_finished_value(raw_value)
         if raw_value.is_a?(::Promise)
-          raw_value.then { |result| super(result) }
+          raw_value.then(->(result) { super(result) }, lambda do |error|
+            raise error unless error.is_a?(GraphQL::ExecutionError)
+            error.ast_node = ast_node
+            query.context.errors << error
+            nil
+          end)
         else
           super
         end
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ class RecordLoader < GraphQL::Batch::Loader
   end
 
   def perform(ids)
-    @model.find(ids).each { |record| fulfill(record.id, record) }
+    @model.where(id: ids).each { |record| fulfill(record.id, record) }
+    ids.each { |id| fulfill(id, nil) unless fulfilled?(id) }
   end
 end
 ```
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -3,6 +3,7 @@ require "promise.rb"
 
 module GraphQL
   module Batch
+    BrokenPromiseError = Class.new(StandardError)
   end
 end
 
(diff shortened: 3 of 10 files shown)


lib/graphql/batch/execution_strategy.rb
```diff
@@ -49,7 +49,13 @@ def each_promise_in_entry(obj, key, value, &block)
     class FieldResolution < GraphQL::Query::SerialExecution::FieldResolution
       def get_finished_value(raw_value)
         if raw_value.is_a?(::Promise)
-          raw_value.then { |result| super(result) }
+          on_reject = lambda do |error|
+            raise error unless error.is_a?(GraphQL::ExecutionError)
+            error.ast_node = ast_node
+            query.context.errors << error
```
dylanahsmith: Perhaps this could be done in the upstream get_finished_value if a GraphQL::ExecutionError instance is passed into it instead of doing this in the result method upstream.

eapache: Yes, it should be a pretty simple refactor once https://github.com/rmosolgo/graphql-ruby/pull/94 is accepted

lib/graphql/batch/execution_strategy.rb
```diff
@@ -49,7 +49,13 @@ def each_promise_in_entry(obj, key, value, &block)
...
+            nil
+          end
+          raw_value.then { |result| super(result) }.then(nil, on_reject)
```
eapache: calling `then` twice when it explicitly has a two-argument form feels a bit weird, as does defining `on_reject` as a variable when it is used only once, however those are both just style things, YMMV

eapache:
:+1:

lib/graphql/batch/execution_strategy.rb
dylanahsmith: I guess we don't really have to worry about GraphQL::ExecutionError being raised from get_finished_value, so I can probably use the single `.then`

I think having a `.rescue` or `.catch` in promise.rb would make it more readable to specify an on_reject callback.

dylanahsmith: changed it to

```ruby
          raw_value.then(->(result) { super(result) }, lambda do |error|
            raise error unless error.is_a?(GraphQL::ExecutionError)
            error.ast_node = ast_node
            query.context.errors << error
            nil
          end)
```

eapache: :+1: to a `rescue` or `catch` method on the promise.

Note that once https://github.com/rmosolgo/graphql-ruby/pull/94 lands this block should return the error (instead of nil) so that the gem can handle errors in non-null fields appropriately.

PR #7, opened 2016-03-25
Provide a mutation execution strategy to sync between top-level fields

dylanahsmith:
@eapache for review

## Problem

Loaders could get batched across top-level mutation fields, which are supposed to be executed serially.

## Solution

Provide a MutationExecutionStrategy with a FieldResolution that waits for promises to resolve after the field is resolved.  It also switches to the batch ExecutionStrategy to resolve sub-selections followed by execution of all of its batch loads.


README.md
```diff
@@ -48,7 +48,7 @@ Use the batch execution strategy with your schema
 ```ruby
 MySchema = GraphQL::Schema.new(query: MyQueryType)
 MySchema.query_execution_strategy = GraphQL::Batch::ExecutionStrategy
```
eapache: perhaps this one should then be called `QueryExecutionStrategy`

eapache: Or I guess not, since we're still dispatching into it for mutation sub-query resolution.

lib/graphql/batch/mutation_execution_strategy.rb
```diff
@@ -0,0 +1,19 @@
...
+        context = execution_context.query.context
+        old_execution_strategy = context.execution_strategy
+        begin
+          context.execution_strategy = GraphQL::Batch::ExecutionStrategy.new
+          raw_value = GraphQL::Batch::Promise.resolve(raw_value).sync
+          result = super(raw_value)
+          GraphQL::Batch::Executor.current.wait_all
+          result
+        ensure
+          old_execution_strategy = context.execution_strategy
```
eapache: should this assignment not be the other way around?

lib/graphql/batch/mutation_execution_strategy.rb
```diff
@@ -0,0 +1,19 @@
+module GraphQL::Batch
+  class MutationExecutionStrategy < GraphQL::Query::SerialExecution
+    class FieldResolution < GraphQL::Query::SerialExecution::FieldResolution
+      def get_finished_value(raw_value)
+        context = execution_context.query.context
+        old_execution_strategy = context.execution_strategy
+        begin
+          context.execution_strategy = GraphQL::Batch::ExecutionStrategy.new
+          raw_value = GraphQL::Batch::Promise.resolve(raw_value).sync
```
eapache: IIUC this line can go above the setting of the execution strategy and therefore outside the `begin/ensure` block; I think I'd prefer that just to make it super-obvious which lines need the new execution strategy (specifically, the `super` call)

dylanahsmith: You're right.  I've moved this line out of the begin/ensure block.

eapache:
LGTM

PR #11, opened 2016-05-31
Add support for checking for unbatched queries by checking a loading flag

dylanahsmith:
@eapache for review

## Problem

When depending on GraphQL::Batch for avoiding N+1 queries, it becomes important to test for unbatched queries.

I have managed to accomplish that by overriding GraphQL::Field to separate the resolver into a preload_proc and a resolve_proc, so the preload_proc can do the batch loading then we can assert no queries are performed in the resolve_proc.  However, this required the result of the preload_proc to be memoized so it could be used in the resolve_proc and also doesn't test the normal code path for executing a graphql query.

## Solution

Add a GraphQL::Batch::Executor#loading attribute which is set to `true` when performing a batch load and `false` otherwise.

This could be used in an `ActiveSupport::Notifications.subscribe` block to detect queries that are performed while GraphQL::Batch::Executor#loading is `false`, even in an integration test.

### DIFF
--- a/lib/graphql/batch/executor.rb
+++ b/lib/graphql/batch/executor.rb
@@ -9,8 +9,14 @@ module GraphQL::Batch
 
     attr_reader :loaders
 
+    # Set to true when performing a batch query, otherwise, it is false.
+    #
+    # Can be used to detect unbatched queries in an ActiveSupport::Notifications.subscribe block.
+    attr_reader :loading
+
     def initialize
       @loaders = {}
+      @loading = false
     end
 
     def shift
@@ -18,7 +24,7 @@ module GraphQL::Batch
     end
 
     def tick
-      shift.resolve
+      with_loading(true) { shift.resolve }
     end
 
     def wait(promise)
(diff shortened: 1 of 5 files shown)


lib/graphql/batch/executor.rb
```diff
@@ -35,5 +36,21 @@ def wait_all
     def clear
       loaders.clear
     end
+
+    def defer
```
eapache: what is this for?

eapache:
One question but LGTM

lib/graphql/batch/executor.rb
dylanahsmith: promise.rb defines it to just yield so that it can be overridden by a derived class to actually defer callbacks.  E.g. you could defer the callback to run on next tick when using EventMachine.  I'm just using this hook to set `loading` to `false` in the callback.  I would prefer not to defer calling the callback though, since that results in backtraces that are missing the context to debug exceptions.

eapache: 🆗

dylanahsmith: Note that if someone wanted to override this method and actually defer the callback, then the callback would get called outside of the parent `with_loading` block.

I'll add some comments to the code to clarify the purpose of `loading` outside this PR, and mention why `defer` is overridden here.

PR #12, opened 2016-06-07
Fix errant `Graphql` constant

davidcelis:
In `lib/graphql/batch/version`, we're accidentally defining a top-level constant that's unused anywhere else in the ecosystem: `Graphql`. It looks like this probably happened during generating the gem, but was never updated to match the `GraphQL` constant that's used everywhere else in this gem and other gems.


dylanahsmith:
yup, that is exactly what happened.  Thanks for fixing it

PR #13, opened 2016-07-06
Also group loaders by their class

dylanahsmith:
@eapache for review

## Problem

Loaders that take the same grouping arguments to the `for` method can cause conflicts.  This becomes obvious when working with singleton loaders, loaders with no grouping arguments passed to `for`.  In these cases, loads from different loaders will get grouped together and performed using the loader instance for the first class used for batch loading.

## Solution

Prepend the loader class to the grouping arguments to use for the loader key.


eapache:
👍

PR #14, opened 2016-08-08
Nicer syntax for parameterless executors

sj26:
Not all loaders require parameters, and using this is a bit awkward:

    promise = SpecialLoader.for.load(1)

This is a bit nicer:

    promise = SpecialLoader.load(1)

After this change, tests still passing:

```
Finished in 0.017371s, 1669.4325 runs/s, 2302.6655 assertions/s.
29 runs, 40 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -5,6 +5,14 @@ module GraphQL::Batch
       Executor.current.loaders[loader_key] ||= new(*group_args)
     end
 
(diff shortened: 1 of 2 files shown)


dylanahsmith:
Yes, we have ended up defining these class methods where no grouping arguments are needed.  I guess it wouldn't hurt to just define them in the base loader class.

PR #15, opened 2016-09-09
Add explicitly check for promise key presence in fulfillment methods, error message

filp:
Right now if there's a mismatch in promise keys, through programmer mistake, a call to `#fulfill` and `#fulfilled?` will be made on `nil`. For example:

```ruby
class SomeLoader
  # ...
  def load(ids)
    # ids is an array of strings, which ActiveRecord happily works with

    # when fulfilling the promise, I use the #id on the record, an int
    @model.where(id: ids).each { |record| fulfill(record.id, record) }
  end
end
```

And then in `lib/graphql/batch/loader.rb`
```ruby
def fulfill(key, value)
  # promises_by_key[key] #=> nil
  promises_by_key[key].fulfill(value)
end
```

While this mistake was my fault alone, it seems like it'd be a common enough case (specially in the early stages of a project) where an explicit check and accompanying error is warranted.

I was only going to open an issue, but decided to go ahead and add some code to match, while trying to change as little as possible of what was already in place.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -30,10 +30,12 @@ module GraphQL::Batch
     end
 
     def fulfill(key, value)
+      expect_promise_key!(key)
       promises_by_key[key].fulfill(value)
     end
 
     def fulfilled?(key)
+      expect_promise_key!(key)
       promises_by_key[key].fulfilled?
     end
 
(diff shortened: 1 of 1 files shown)


dylanahsmith:
Thanks for pointing out this problem.  I decided to just use Array#fetch so that it provides a KeyError and to provide the key name in the error message.

PR #19, opened 2016-10-06
Add missing syntax highlight in README

jtdowney:
This is very minor but I noticed it while reading through the project README.


dylanahsmith:
Thanks

PR #20, opened 2016-10-11
Add regression test for execution errors in non-nullable fields

### LINKED ISSUE #18: Non-nullable fields that raise an ExecutionError cause an uncaught exception
While investigating an issue I noticed that when using `graphql-batch`, a non-nullable field that raises a `GraphQL::ExecutionError` causes an uncaught `GraphQL::InvalidNullError` exception.

I have yet to find the exact commit in `graphql-ruby` that caused this regression, but it's reproducible with the following diff:

```diff
diff --git a/test/batch_test.rb b/test/batch_test.rb
index daf48f0..4234c92 100644
--- a/test/batch_test.rb
+++ b/test/batch_test.rb
@@ -79,6 +79,22 @@ class GraphQL::BatchTest < Minitest::Test
     assert_equal ["Product/123"], queries
   end

+  def test_non_null_that_raises
+    query_string = <<-GRAPHQL
+      {
+        product(id: "1") {
+          id
+          nonNullButRaises {
+            id
+          }
+        }
+      }
+    GRAPHQL
+    result = Schema.execute(query_string)
+  end
+
   def test_batched_association_preload
     query_string = <<-GRAPHQL
       {
diff --git a/test/support/schema.rb b/test/support/schema.rb
index c6eb843..aa97340 100644
--- a/test/support/schema.rb
+++ b/test/support/schema.rb
@@ -55,6 +55,13 @@ ProductType = GraphQL::ObjectType.define do
       Promise.all([query]).then { query.value.size }
     }
   end
+
+  field :nonNullButRaises do
+    type !ImageType
+    resolve -> (_, _, _) {
+      raise GraphQL::ExecutionError, 'Error'
+    }
+  end
```

At first I thought it was `graphql-ruby` that didn't handle this case correctly, but then I tried it out and it worked as expected.

I believe in this case you'd expect `result` to be something along the lines of:

```
{
  "data": {
    "products": null
  },
  "errors": [
    ... 
  ]
}
```

@dylanahsmith any idea? If not, I can take a look later this week.

cjoudrey:
re: rmosolgo/graphql-ruby#296 and closes #18 

This adds two regression tests for non-nullable fields that raise.

:eyes: @dylanahsmith

### DIFF
--- a/test/batch_test.rb
+++ b/test/batch_test.rb
@@ -79,6 +79,33 @@ class GraphQL::BatchTest < Minitest::Test
     assert_equal ["Product/123"], queries
   end
 
+  def test_non_null_field_that_raises_on_nullable_parent
+    query_string = <<-GRAPHQL
+      {
+        product(id: "1") {
+          id
+          nonNullButRaises
+        }
+      }
+    GRAPHQL
+    result = Schema.execute(query_string)
+    expected = { 'data' => { 'product' => nil }, 'errors' => [{ 'message' => 'Error', 'locations' => [{ 'line' => 4, 'column' => 11 }], 'path' => ['product', 'nonNullButRaises'] }] }
+    assert_equal expected, result
+  end
+
+  def test_non_null_field_that_raises_on_query_root
+    query_string = <<-GRAPHQL
+      {
+        nonNullButRaises {
+          id
+        }
+      }
+    GRAPHQL
+    result = Schema.execute(query_string)
+    expected = { 'data' => nil, 'errors' => [{ 'message' => 'Error', 'locations' => [{ 'line' => 2, 'column' => 9 }], 'path' => ['nonNullButRaises'] }] }
+    assert_equal expected, result
+  end
+
   def test_batched_association_preload
     query_string = <<-GRAPHQL
       {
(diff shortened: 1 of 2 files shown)


test/batch_test.rb
```diff
@@ -79,6 +79,33 @@ def test_record_missing
     assert_equal ["Product/123"], queries
   end
 
+  def test_non_null_field_that_raises_on_nullable_parent
```
cjoudrey: This was broken in `0.19.1`.

test/batch_test.rb
```diff
@@ -79,6 +79,33 @@ def test_record_missing
...
+          nonNullButRaises
+        }
+      }
+    GRAPHQL
+    result = Schema.execute(query_string)
+    expected = { 'data' => { 'product' => nil }, 'errors' => [{ 'message' => 'Error', 'locations' => [{ 'line' => 4, 'column' => 11 }], 'path' => ['product', 'nonNullButRaises'] }] }
+    assert_equal expected, result
+  end
+
+  def test_non_null_field_that_raises_on_query_root
```
cjoudrey: This was worked in `0.19.1` but added it as a precaution.

PR #22, opened 2016-10-14
Provide a useful error when load is called after perform

### LINKED ISSUE #21: Cannot re-use loaders
I was trying to reuse a loader, but if the second load provides a different input id it will crash with `GraphQL::Batch::BrokenPromiseError: Promise wasn't fulfilled after all queries were loaded`.

Example:

```ruby
loader = SomeLoader.for
loader.load(1).sync # First load
loader.load(1).sync # Everything's okay. Record already loaded.

loader.load(2).sync # Crashes.
# gems/promise.rb-0.7.1/lib/promise.rb:84:in `block in reject'
# gems/promise.rb-0.7.1/lib/promise.rb:114:in `dispatch'
# gems/promise.rb-0.7.1/lib/promise.rb:82:in `reject'
# gems/graphql-batch-0.2.4/lib/graphql/batch/executor.rb:33:in `wait'
# gems/graphql-batch-0.2.4/lib/graphql/batch/promise.rb:4:in `wait'
# gems/promise.rb-0.7.1/lib/promise.rb:64:in `sync'
```

It crashes before reaching the `#perform` method at all.

Is this by design? I think the README should mention it in that case.

dylanahsmith:
Fixes #21

@Mange what do you think about a runtime error of "Loader can't be used after batch load" to avoid confusion?

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -22,6 +22,9 @@ module GraphQL::Batch
     end
 
     def load(key)
+      if promises_by_key.frozen?
+        raise "Loader can't be used after batch load"
+      end
       promises_by_key[key] ||= Promise.new
     end
 
@@ -43,6 +46,7 @@ module GraphQL::Batch
     end
 
     def resolve
+      promises_by_key.freeze
       perform(keys)
       check_for_broken_promises
     rescue => err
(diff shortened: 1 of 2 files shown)


Mange:
This makes sense IMO. I also note that you cannot access already loaded instances anymore, but I think the consistency is good[1]. It removes a lot of surprises. ☺️

[1]: Well, and the fact that a "loader" shouldn't act as a repository. If one wants a lazy loaded repository, it should be composed *with* loaders, not *be* a loader.

Issue #24, opened 2016-10-21
Document how one could preload has_many associations


cjoudrey:
There was some discussion about how one might preload `has_many` associations in `#graphql`.

It might be interesting to share some of our `AssociationLoader` logic or writing a quick tutorial to achieve this in the `README`.

dylanahsmith:
I think we should make our DerivedKeyLoader builtin functionality, since is something we use to properly implement this in Shopify

nickpoorman:
Has anyone written this up? Would love to see how you did it.

sjchmiela:
I created for my own purposes an `ArrayLoader`, which you use providing a value and a key of a has-many association, like this:

`ArrayLoader.for(Comment).load([article.id, :article_id])`

```Ruby
class ArrayLoader < GraphQL::Batch::Loader
  def initialize(model, default = nil)
    @model = model
    @default = default
  end

  def perform(keys)
    queries = {}
    keys.each do |key|
      if key.is_a?(Array)
        query_value, query_key = key
        queries[query_key] = [] if queries[query_key].nil?
        queries[query_key] << query_value
      end
    end

    results_cache = {}

    queries.each do |key, values|
      @model.where(key => values).each { |record|
        results_cache[[record[key], key]] = [] if results_cache[[record[key], key]].nil?
        results_cache[[record[key], key]] << record
      }
    end

    results_cache.each do |key, values|
      fulfill(key, values)
    end

    keys.each { |key| fulfill(key, @default || []) unless fulfilled?(key) }
  end
end
```

cainlevy:
This is what I've hit on:

```
class AssociationLoader < GraphQL::Batch::Loader
  def initialize(model, association)
    @model = model
    @association = association
  end

  def perform(vals)
    ActiveRecord::Associations::Preloader.new.preload(vals, @association)
    vals.each { |val| fulfill(val, val.send(@association)) }
  end
end
```

Used like so:

```
AuthorType = GraphQL::ObjectType.define do
  field :books, types[!BookType] do
    resolve ->(obj, _, _) {
      AssociationLoader.for(Author, :books).load(obj)
    }
  end
end

BookType = GraphQL::ObjectType.define do
  field :author, types[!AuthorType] do
    resolve ->(obj, _, _) {
      AssociationLoader.for(Book, :author).load(obj)
    }
  end
end
```

dylanahsmith:
Here is what we are using, which is a bit longer to:
* give better errors if the association doesn't exist
[...]

PR #25, opened 2016-10-21
Simplify the implementation of a derived cache key loader

dylanahsmith:
@cjoudrey & @xuorig please review

This will make #24 easier, because in Shopify we use a DerivedKeyLoader to implement our AssociationLoader:

```ruby
  module DerivedKeyLoader
    # GraphQL::Batch::Loader helper for when the key by which you want to identify your records
    # does not in itself contain enough information to implement `perform`

    def load(record)
      key = key_for_record(record)
      records_by_key[key] ||= record
      super(key)
    end

    def keys
      records_by_key.values_at(*super)
    end

    def fulfill(record, value)
      super(key_for_record(record), value)
    end

    private

    def records_by_key
      @records_by_key ||= {}
    end

    def key_for_record(record)
      # by default use the record's object id, this is a nice sane default and works well for association loaders
      # (the initial users of this code) who want to load their association on all extant record objects sharing the
      # same database id
      record.object_id
    end
  end
```

Now that can be simplified to just overriding GraphQL::Batch::Loader#cache_key in the AssociationLoader

```ruby
  def cache_key(load_key)
    load_key.object_id
  end
```

I was able to make such a simplification by avoiding the need to do the reverse lookup of the load key from the cache key.  Instead, we store the load keys to pass to Loader#perform in an array, separate from the hash that maps cache_key => promise.

This is also a precursor to being able to re-use a loader to return resolved promises after they have been loaded, which is why I named the hash `cache` and the array of load keys as `queue`.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -13,55 +13,78 @@ module GraphQL::Batch
       self.for.load_many(keys)
     end
 
-    def promises_by_key
-      @promises_by_key ||= {}
-    end
-
-    def keys
-      promises_by_key.keys
-    end
-
     def load(key)
-      if promises_by_key.frozen?
+      if queue.frozen?
         raise "Loader can't be used after batch load"
       end
-      promises_by_key[key] ||= Promise.new
+      cache[cache_key(key)] ||= begin
+        queue << key
+        Promise.new
+      end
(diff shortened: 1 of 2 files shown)


xuorig:
lgtm! 🚢

PR #29, opened 2016-10-31
Allow loaders to be re-used so they can cache loads.

dylanahsmith:
@cjoudrey & @xuorig please review
cc @Mange since you opened #21

This allows a loader to be re-used so that it can be used to cache previous loads.

Currently this will require keeping a reference to the loader to re-use it.  A follow-up will be needed to have the executor keep references to the old loaders during a GraphQL request.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -2,7 +2,9 @@ module GraphQL::Batch
   class Loader
     def self.for(*group_args)
       loader_key = [self].concat(group_args)
-      Executor.current.loaders[loader_key] ||= new(*group_args)
+      Executor.current.loaders[loader_key] ||= new(*group_args).tap do |loader|
+        loader.loader_key = loader_key
+      end
     end
 
     def self.load(key)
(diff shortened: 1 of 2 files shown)


Mange:
Thanks for the CC. Looks like a great addition!

lib/graphql/batch/loader.rb
```diff
@@ -13,9 +15,12 @@ def self.load_many(keys)
...
     end
 
+    attr_accessor :loader_key
+
     def load(key)
-      if queue.frozen?
-        raise "Loader can't be used after batch load"
+      loader = Executor.current.loaders[loader_key] ||= self
+      if loader != self
+        raise "load called on loader that wasn't registered with executor"
```
cjoudrey: Just to make sure I understand this, this would occur if you call `a = Loader.for(:something)` and then use `a` from either another thread? Would this also raise across different queries on a single Thread?

dylanahsmith: It could happen if you are mixing up re-using a loader and getting a loader using Loader.for, which could happen from using a loader in another thread, but could also happen in a single thread:

```ruby
  def test_load_on_different_loader
    loader = EchoLoader.for
    assert_equal :a, loader.load(:a).sync
    loader2 = EchoLoader.for
    promise = loader2.load(:b)
    err = assert_raises(RuntimeError) do
      loader.load(:c).sync
    end
    assert_equal "load called on loader that wasn't registered with executor", err.message
  end
```

PR #33, opened 2016-11-15
Lazy resolution API

rmosolgo:
(Not quite ready: needs graphql 1.3) 

GraphQL 1.3 will include a `lazy_resolve` API which borrows __heavily__ the from `GraphQL::Batch::ExecutionStrategy` (It's a total ripoff! Thank you so much for figuring this out long ago!). I recommend it for a few reasons: 

- No more maintenance issues with regard to refactoring `SerialExecution` 
- `execution_strategy` is not a good extension point. At least, it's not good for adding features to GraphQL, because you can only use one at a time. If I implemented `@defer` as an execution strategy (I won't), you'd have to _choose_, batching or `@defer`? I'd like to allow them to be used together. I made a mistake on `execution_strategy` and I'd like to improve on that!
- Slightly improved null handling. The new behavior here is described in 88e5d92 -- basically, nulls don't propagate by `raise`-ing through the Ruby stack anymore, so you can correctly "backtrack" through the response instead of nullifying everything. 

What do you think? How does this look to you? 

Thanks again for figuring this out first! Continuing to support graphql-batch is a must-have for me because this project is what makes graphql-ruby _actually_ viable!


rmosolgo:
(oh right, tests don't pass because the new version isn't on rubygems yet!)

test/support/schema.rb
```diff
@@ -151,6 +163,5 @@
 Schema = GraphQL::Schema.define do
   query QueryType
   mutation MutationType
-  query_execution_strategy GraphQL::Batch::ExecutionStrategy
-  mutation_execution_strategy GraphQL::Batch::MutationExecutionStrategy
+  lazy_resolve(GraphQL::Batch::Promise, :sync)
```
dylanahsmith: I'll make a promise.rb release with https://github.com/lgierth/promise.rb/pull/24 so `lazy_resolve(Promise, :sync)` can be used here and Promise.all can be used in the resolvers.  This does make sense without that dependancy bump on promise.rb though

lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,66 +0,0 @@
...
-  class ExecutionStrategy < GraphQL::Query::SerialExecution
-    attr_accessor :disable_batching
-
-    def execute(_, _, query)
-      as_promise(super).sync
-    rescue GraphQL::InvalidNullError => err
-      err.parent_error? || query.context.errors.push(err)
-      nil
-    ensure
-      GraphQL::Batch::Executor.current.clear
```
dylanahsmith: Looks like we might need to caution users about how state can now leak between requests without an ensure block like this.

For example, if some loader doesn't get loaded, then that load could happen on a following query.  Also, since loads a batched, it is possible for a specific key to cause a loader to raise an exception, causing a following request to fail from a load in a previous request.

rmosolgo: Oh yikes, that's bad! Would any of these approaches work to save users from thinking about it? 

- Clear the executor _before_ starting a query 
- Provide a `begin ... ensure` wrapper for user code like `GraphQL::Batch.isolate { ... }` (but with a human-friendly name) 
- `instrument(:query, GraphQL::Batch::QueryInstrumentation)` to ensure cleanup (this doesn't work yet -- I bet the after-query hook isn't called if there's an error, but that could be fixed)

rmosolgo:
I see you have an up-to-date branch that includes instrumentation AND backwards compatibility, that looks better than this one!

https://github.com/Shopify/graphql-batch/tree/lazy-resolve-api

PR #35, opened 2016-11-18
Allow loaders to be used without an executor

dylanahsmith:
@rmosolgo, @xuorig & @cjoudrey 

## Problem

As pointed out in https://github.com/Shopify/graphql-batch/pull/33#discussion_r88349952, when using graphql-batch outside of the execution strategy, we run the risk of leaky state.  This could be a problem when using the graphql gem's new lazy resolution API, but it could also be a problem in tests.

As such, I think we should move away from depending on global state, even if we support one for convenience and backwards compatibility.

The thread local executor is used for three purposes right now:
1. stores a list of loaders
2. executes those loaders
3. keeps track of the loading flag that unit tests can check to catch unbatched queries

## Solution

promise.rb 0.7.2 now keeps track of the source of the promise and uses that to sync the promise by default.  All that is needed to use that is to set the promise source on the promise returned by the loader's load method.

Without the need to use the executor for resolve the promises, we could then create per request loaders as documented in the [dataloader README](https://github.com/facebook/dataloader#creating-a-new-dataloader-per-request) which could then be stored in the graphql query context.  Doing this also allows loaders' cache to be used after their batch load is performed.  However, this does bring the concern of cache invalidation if the loaders are used across top-level mutation fields.

For now the loading flag is kept on the executor and is just not updated when there is no current executor.  That could be fixed in the future by moving the flag outside of the executor into it's own thread-local variable, but will more likely be replaced by a more generic instrumentation hook.

## Follow-up

I plan on having the executor not get created automatically and require the code using the library to specify where batching starts and stops using a block.  E.g.

```ruby
product = GraphQL::Batch.batch do
    RecordLoader.for(Product).load(product_id)
end
```

would setup the executor for the block and pass the result of the block to `Promise.sync`

### DIFF
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -1,12 +1,16 @@
 module GraphQL::Batch
   class ExecutionStrategy < GraphQL::Query::SerialExecution
     def execute(_, _, query)
-      Promise.sync(as_promise_unless_resolved(super))
+      deep_sync(super)
     rescue GraphQL::InvalidNullError => err
       err.parent_error? || query.context.errors.push(err)
       nil
     ensure
       GraphQL::Batch::Executor.current.clear
+    end
+
+    def deep_sync(result)
+      Promise.sync(as_promise_unless_resolved(result))
     end
 
     private
--- a/lib/graphql/batch/execution_strategy.rb
+++ b/lib/graphql/batch/execution_strategy.rb
@@ -21,7 +21,7 @@ module GraphQL::Batch
         end
       end
       return result if all_promises.empty?
-      Promise.all(all_promises).then { result }
+      ::Promise.all(all_promises).then { result }
     end
 
     def each_promise(obj, &block)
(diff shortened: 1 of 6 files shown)


lib/graphql/batch/execution_strategy.rb
```diff
@@ -1,14 +1,18 @@
...
+      deep_sync(super)
     rescue GraphQL::InvalidNullError => err
       err.parent_error? || query.context.errors.push(err)
       nil
     ensure
       GraphQL::Batch::Executor.current.clear
     end
 
+    def deep_sync(result)
+      Promise.sync(as_promise_unless_resolved(result))
```
xuorig: do you still need `as_promise_unless_resolved` with `Promise.sync` ?

dylanahsmith: `as_promise_unless_resolved` recursively looks for promises in the result and wraps the whole result in a promise if it finds any.  Promise.sync only checks if the object itself is a promise

rmosolgo:
:+1: I think I see, basically if you `.wait` a promise which was created without an `executor.current`, it will resolve that promise's loader (its `source`)

Issue #27, opened 2016-10-28
Instrumentation


tmeasday:
Following up from https://github.com/rmosolgo/graphql-ruby/pull/354#issuecomment-256810751:

@dylanahsmith : 

> For graphql-batch I would just recommend using Promise.map_value to get the end of the field resolution:

```ruby
class MyFieldInstrumentation
  def instrument(type, field)
    old_resolve_proc = field.resolve_proc
    new_resolve_proc = ->(obj, args, ctx) {
      Rails.logger.info("#{type.name}.#{field.name} START: #{Time.now.to_i}")
      result = old_resolve_proc.call(obj, args, ctx)
      Promise.map_value(result) do |value|
        Rails.logger.info("#{type.name}.#{field.name} END: #{Time.now.to_i}")
        value
      end
    }

    field.redefine do
      resolve(new_resolve_proc)
    end
  end
end
```

Actually the above does not work (or at least not in a particularly useful way). The problem is that the executor does not necessarily run the loader associated with the field immediately, instead it may wait for some other loaders (or fields I suppose) to execute first. So the "start time" logged above (when the field is added to the loader), is not really "correct", in terms of the work done to fetch the field.

As an example, if I run a query against our [example GitHunt server](https://github.com/apollostack/githunt-api-rails) that looks something like:

```
{
  entries {
     # this field goes to a batch loader that talks to github
     postedBy
     # this field goes to another github batch loader
     repository
     # this field goes to a simple SQL batch loader
     votes
  }
}
```

You end up seeing traces that look like:

<img width="837" alt="screenshot 2016-10-28 08 54 09" src="https://cloud.githubusercontent.com/assets/132554/19813021/b1c1003e-9cec-11e6-9b69-5f39ef2cc198.png">

Note in the above that the start time of all the fields is more or less the same (as the executor runs over the set of entries, and they are all added to the various loaders), and the total time is "cumulative". In actuality the `votes` loader is more or less instantaneous (running against a local sqlite db in this case), and a correct start time for those fields should really be at the *end* of the `repository` loader (so I guess 2.03ms in this SS).

dylanahsmith:
Yes, that code example tracks the start and end of resolving a field, not the time it spends actually executing those fields, so no other integration is needed with graphql-ruby.  You would see a similar problem any time fields are able to resolved completely concurrently, such as from thread/connection pools, locking or CPU intensive tasks in a parallel executor.

The problem really is about instrumenting batch loaders.

[...]

PR #36, opened 2016-11-18
Require loaders to be used within a GraphQL::Batch.batch block

dylanahsmith:
@rmosolgo, @xuorig & @cjoudrey

## Problem

As pointed out in https://github.com/Shopify/graphql-batch/pull/33#discussion_r88349952 and https://github.com/Shopify/graphql-batch/pull/35, by storing state in thread local variables we run the risk of state leaking between requests or tests.

## Solution

Change GraphQL::Batch::Executor.current so that it no longer automatically creates an executor so that the scope of batching is done explicitly using `GraphQL::Batch.batch`.  E.g.

```ruby
product = GraphQL::Batch.batch do
    RecordLoader.for(Product).load(product_id)
end
```

will assign an executor to GraphQL::Batch::Executor.current, call Promise.sync on the result of the block and clear the executor at the end of the block.

The reason why `Promise.sync` is needed on the result is because the promise callbacks might also do batch loads using GraphQL::Batch::Loader.for to get the loader from GraphQL::Batch::Executor.current.


rmosolgo:
:+1: In a world with no `ExecutionStrategy`, how do you think this could be hooked up to `Schema#execute`?

dylanahsmith:
I was thinking it could be called outside of graphql-ruby around the call to `Schema#execute`.  There is a before_query and after_query instrumentation in graphql-ruby, but an exception would prevent the after_query method from being called.

Do you have any suggestions for how you would like it to integrate with graphql-ruby?

rmosolgo:
:+1: for instrumentation, I think this `ensure` would do the trick: https://github.com/rmosolgo/graphql-ruby/pull/412


Then you could have 

```ruby 
GraphQL::Schema.define do 
  # ... 
  instrument(:query, GraphQL::Batch::Intstrumentation) 
end 
``` 

I'd be happy to update my open PR with that code if it looks ok to you!

dylanahsmith:
That sounds good, except the name `GraphQL::Batch::Instrumentation` sounds like something for https://github.com/Shopify/graphql-batch/issues/27 rather than something to start and end the batch.  What do you think about `instrument(:query, GraphQL::Batch::Setup)` to make its purpose clearer?

rmosolgo:
:+1: for `Setup` !

dylanahsmith:
> I'd be happy to update my open PR with that code if it looks ok to you!

:+1:

PR #37, opened 2016-11-22
Support graphql-ruby lazy resolution API

dylanahsmith:
@rmosolgo
cc @cjoudrey & @xuorig 

This should support both the new lazy resolution API that will come in graphql-ruby 1.3.0 as well as the old execution strategy, while also not breaking when the execution strategies are removed.

Nothing in this PR is specific to graphql-ruby 1.3.0 other than test/lazy_resolve_test.rb which uses feature detection to only run the test if the lazy_resolve method is defined, so we should be able to make a graphql-batch 0.3.0 release without waiting for graphql-ruby 1.3.0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 /.bundle/
 /.yardoc
-/Gemfile.lock
+/Gemfile*.lock
 /_yardoc/
 /coverage/
 /doc/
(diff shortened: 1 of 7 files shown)


rmosolgo:
:+1: 

For `graphql 1.3.0`, I was only waiting to make sure it would have everything needed to support this change. It looks like it does, so I could cut it tonight!

Issue #40, opened 2017-01-05
Changelog


jsdnxx:
I just saw that version 0.3 was released- congratulations! Is there a list of what has changed? I would recommend also checking it into the repo in a changelog file.

Cheers

dylanahsmith:
https://github.com/Shopify/graphql-batch/compare/v0.2.5...v0.3.0 for a full list of changes.  https://github.com/Shopify/graphql-batch/pull/36 is a breaking change but should only affect test code.  The main feature is support for https://github.com/rmosolgo/graphql-ruby/blob/master/guides/schema/lazy_execution.md which was released after the last release of this library, so I should update the README to recommend the use of that feature instead of the graphql-batch execution strategies.

sjchmiela:
> recommend the use of that feature instead of the graphql-batch execution strategies.

and `graphql-ruby` recommends use of graphql-batch…

> The example above is simple and has some shortcomings. Consider the graphql-batch gem for a robust solution to batched resolution.

Which should one use then?

dylanahsmith:
I'm not saying to stop using graphql-batch, just that instead of doing

```ruby
MySchema = GraphQL::Schema.define do
  query MyQueryType
end
MySchema.query_execution_strategy = GraphQL::Batch::ExecutionStrategy
MySchema.mutation_execution_strategy = GraphQL::Batch::MutationExecutionStrategy
```

you would do

```ruby
MySchema = GraphQL::Schema.define do
  query MyQueryType
  lazy_resolve(Promise, :sync)
  instrument(:query, GraphQL::Batch::Setup)
end
```

GraphQL::Batch::ExecutionStrategy and GraphQL::Batch::MutationExecutionStrategy will be deprecated then removed since they rely on method overrides which make them fragile to changes in graphql-ruby

sjchmiela:
Ah, all right, that I couldn't even think of. :)

jsdnxx:
Ah, I didn't even think to use GitHub's compare :) Thanks!

PR #41, opened 2017-01-06
Update README.md

sjchmiela:
https://github.com/Shopify/graphql-batch/issues/40#issuecomment-270868706

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,7 @@ class RecordLoader < GraphQL::Batch::Loader
 end
 ```
 
-Use the batch execution strategy with your schema
-
-- When running `graphql` gem versions `>= 1.3.0`:
+Use lazy execution and instrumentation by `GraphQL::Batch` in your schema
 
 ```ruby
 MySchema = GraphQL::Schema.define do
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,6 +48,19 @@ end
...
+```ruby
+MySchema = GraphQL::Schema.define do
+  query MyQueryType
+  
+  lazy_resolve(Promise, :sync)
+  instrument(:query, GraphQL::Batch::Setup)
+end
+```
+
+- When running `graphql` gem versions `< 1.3.0`:
```
dylanahsmith: I would prefer to just recommend using lazy_resolve and deprecating the old way of doing it.

I have opened https://github.com/Shopify/graphql-batch/pull/42 to deprecate the old way, although I think I will need to also add a warning for when an old version of the graphql gem is used.

README.md
```diff
@@ -48,6 +48,19 @@ end
 
 Use the batch execution strategy with your schema
```
dylanahsmith: This sentence is describing the old way changing the execution strategy, which isn't applicable to the lazy_resolve approach.

README.md
sjchmiela: Removed deprecated way

README.md
sjchmiela: Changed to:
> Use lazy execution and instrumentation by `GraphQL::Batch` in your schema

PR #49, opened 2017-02-23
Expose `use_graphql_batch` instead of internal setup instructions

cjoudrey:
👋 @dylanahsmith @xuorig @rmosolgo

The setup instruction for `graphql-batch` kind of bugged me.

It felt as if we were exposing implementation details. Also, what happens if we ever need to change these instructions for whatever reason? It would be a shame to have to include "upgrade instructions" in the `README.md`.

I propose we hide these setup methods in the library and only expose an idempotent `use_graphql_batch` method in the `GraphQL::Schema` DSL.

Thoughts?

### DIFF
--- a/lib/graphql/batch/setup.rb
+++ b/lib/graphql/batch/setup.rb
@@ -12,3 +12,12 @@ module GraphQL::Batch
     end
   end
 end
+
+GraphQL::Schema.accepts_definitions(
+  use_graphql_batch: ->(schema) {
+    return if schema.metadata[:graphql_batch_setup]
+    schema.instrument(:query, GraphQL::Batch::Setup)
+    schema.lazy_methods.set(Promise, :sync)
+    schema.metadata[:graphql_batch_setup] = true
+  }
+)
--- a/README.md
+++ b/README.md
@@ -52,8 +52,7 @@ Use lazy execution and instrumentation by `GraphQL::Batch` in your schema
 MySchema = GraphQL::Schema.define do
   query MyQueryType
 
-  lazy_resolve(Promise, :sync)
-  instrument(:query, GraphQL::Batch::Setup)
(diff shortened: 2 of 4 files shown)


lib/graphql/batch/setup.rb
```diff
@@ -12,3 +12,12 @@ def after_query(query)
     end
   end
 end
+
+GraphQL::Schema.accepts_definitions(
+  use_graphql_batch: ->(schema) {
+    return if schema.metadata[:graphql_batch_setup]
+    schema.instrument(:query, GraphQL::Batch::Setup)
+    schema.lazy_methods.set(Promise, :sync)
```
rmosolgo: If you're interested in exploring this option, I'd be happy to add `Schema#lazy_resolve(result_class, method_name)` for this case!

rmosolgo:
```ruby 
  GraphQL::Batch.use(self)
```

This is really interesting to me, what if graphql-ruby made this even easier, something like 

```ruby 
Schema.define do 
  use MyPlugin # calls MyPlugin.use(self)
end 
```

?

dylanahsmith:
@rmosolgo that sounds good to me.  It would also make it easy to provide `GraphQL::Batch.use` without depending on that proposed `use` method being available in graphql-ruby.

rmosolgo:
I might go one further and make it 

```ruby 
Schema.define do 
  use MyPlugin, plugin_option: "whatever" # calls MyPlugin.use(self, options = {})
end 
```

Then I can migrate a bunch of graphql-pro features to that API too!

xuorig:
That's very cool @rmosolgo. Let me know if you need a hand 👍

rmosolgo:
✋ I need a hand 😬

cjoudrey:
Closing in favour of #51.

PR #51, opened 2017-04-26
Make GraphQL::Batch a plugin by adding `use`

swalkinshaw:
Adds the `use` method so `GraphQL::Batch` can be used as a plugin in `define` blocks.

This replaces https://github.com/Shopify/graphql-batch/pull/49

/cc @cjoudrey

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,23 @@ class RecordLoader < GraphQL::Batch::Loader
 end
 ```
 
-Use lazy execution and instrumentation by `GraphQL::Batch` in your schema
+Use `GraphQL::Batch` as a plugin in your schema (for graphql >= `1.5.0`).
 
 ```ruby
 MySchema = GraphQL::Schema.define do
   query MyQueryType
 
-  lazy_resolve(Promise, :sync)
-  instrument(:query, GraphQL::Batch::Setup)
+  use GraphQL::Batch
+end
+```
+
+For pre `1.5.0` versions:
+
+```ruby
+MySchema = GraphQL::Schema.define do
+  query MyQueryType
+
+  GraphQL::Batch.use(self)
 end
 ```
 
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -19,6 +19,11 @@ module GraphQL
       end
     end
 
+    def self.use(schema_defn)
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -46,7 +46,17 @@ class RecordLoader < GraphQL::Batch::Loader
...
+
+```ruby
+MySchema = GraphQL::Schema.define do
+  query MyQueryType
+
+  use GraphQL::Batch
+end
+```
+
+Or, use lazy execution and instrumentation by `GraphQL::Batch` in your schema.
```
dylanahsmith: For older versions of graphql-ruby they should just use `GraphQL::Batch.use(self)` instead of `use GraphQL::Batch`

lib/graphql/batch.rb
```diff
@@ -19,6 +19,15 @@ def self.batch
       end
     end
 
+    def self.use(schema_defn)
+      if schema_defn.target.class != GraphQL::Schema
```
dylanahsmith: `target` and `use` were added in the same PR upstream (https://github.com/rmosolgo/graphql-ruby/pull/565) so using it here would prevent us from being able to recommend using `GraphQL::Batch.use(self)` for older versions of graphql-ruby.

If you do want to keep this check, then I would change it to `if Gem::Version.new(GraphQL::VERSION) >= Gem::Version.new('1.5.0') && schema_defn.target.class != GraphQL::Schema`

README.md
```diff
@@ -46,7 +46,17 @@ class RecordLoader < GraphQL::Batch::Loader
 end
 ```
 
-Use lazy execution and instrumentation by `GraphQL::Batch` in your schema
+Use `GraphQL::Batch` as a plugin in your schema (for graphql > `1.5.0`).
```
dylanahsmith: It was added in version 1.5.0, so you should change `>` to `>=`

swalkinshaw: 1.5.0 was yanked (actually so was .1 and .2), should we make it `>= 1.5.3`?

dylanahsmith: I think you could still say `>= 1.5.0`, since it is still true.  Yanking a version doesn't change history.

lib/graphql/batch.rb
swalkinshaw: I've just removed it. I don't think it's too helpful and it will have a fairly description error anyway.

PR #52, opened 2017-05-30
Allow loaders to override the loader key

staugaard:
Allow loaders to override the loader key


dylanahsmith:
I'll look into those test failures.  It looks like [graphql-ruby 1.6.0](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#160-27-may-2017) introduced some breaking changes.

PR #53, opened 2017-05-30
Improve loader reuse

staugaard:
Stop deleting loaders from the cache.

This improves loader reuse.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -40,8 +40,8 @@ module GraphQL::Batch
     end
 
     def resolve #:nodoc:
+      return if resolved?
       load_keys = queue
-      return if load_keys.empty?
       @queue = nil
       perform(load_keys)
       check_for_broken_promises(load_keys)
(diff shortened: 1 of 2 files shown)


lib/graphql/batch/loader.rb
```diff
@@ -54,13 +54,16 @@ def resolve #:nodoc:
...
       if executor
-        executor.loaders.delete(loader_key)
         executor.resolve(self)
       else
         resolve
       end
     end
 
+    def resolved?
+      queue.empty?
```
dylanahsmith: Use `@queue.nil? || @queue.empty?` to avoid unnecessarily creating an empty array.  You can also use the same method in `resolve` to `return if resolved?` before `load_keys = queue` instead of using `return if load_keys.empty?` afterwards.

lib/graphql/batch/executor.rb
```diff
@@ -36,7 +36,7 @@ def tick
     end
 
     def wait_all
-      tick until loaders.empty?
+      tick until loaders.values.all? { |loader| loader.resolved? }
```
dylanahsmith: It doesn't look like this is needed, since Loader#resolve returns early if it is already resolved.

staugaard:
How about this?

staugaard:
FYI. Removing the changes to the executor leaves `Executor#wait_all` in a broken state since it will loop for ever since the loaders cache never becomes empty.

dylanahsmith:
`Executor#wait_all` will empty `Executor#loaders` even though `Promise.sync` doesn't

PR #55, opened 2017-06-06
Support GraphQL 1.6's multiplexed execution

theorygeek:
In GraphQL 1.6, @rmosolgo introduced the `Schema#multiplex` feature 🎉 which lets you execute multiple queries in one go. But it seems that it invokes the query instrumentation once per query, instead of just once, which causes the `NestedError` to be raised. This change removes the `NestedError` behavior.

I'm opening this PR mainly as a discussion. I'm actually not sure what the ramifications are of allowing this nested behavior. But anywho, thoughts?

### DIFF
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -10,6 +10,7 @@ module GraphQL
     class NestedError < StandardError; end
 
     def self.batch
+      # TODO: Should this be allowed now?
       raise NestedError if GraphQL::Batch::Executor.current
       begin
         GraphQL::Batch::Executor.current = GraphQL::Batch::Executor.new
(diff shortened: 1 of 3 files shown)


rmosolgo:
I was thinking this might need a tiny bit more change, that the instrumenter could be adjusted to hook on `instrument(:multiplex, ...)`, that way, the single executor could be shared by all queries.

I'd be happy to take it for a spin this afternoon, I think a proper test calling `Schema#multiplex` would be a good way to ensure that it works properly.

rmosolgo:
Here's another take, what do you think?

https://github.com/Shopify/graphql-batch/pull/57

theorygeek:
@rmosolgo Ah nice, yeah I think that's a better solution than the one I have here.

Issue #43, opened 2017-01-19
Loader cache is short lived


farnoy:
Hi, for a query like this:

```graphql
query {
  company(id: 10) {
    job {
      company {
        id
      }
    }
}
```

I checked that (with my loaders) it will do 3 queries. First to fetch the Company, then Job, then Company again. The last query is cached by ActiveRecord, so it's not a big deal. I am wondering if it should be cached in graphql-batch instead. Both the top level `company` field and `company` under `Job` type are essentially resolved like this: `RecordLoader.for(Company).load(10)` but the caches seem to live until all direct children and siblings are resolved. In other words, the cache is effectively purged by the time my nested `company` field is being resolved.

Is there a reason why we couldn't cache these loaders for longer periods of time?

dylanahsmith:
Yes, it should cache for the whole query.  I have made steps in this direction, but it is still a remaining issue.

Right now it only does batching, but no caching.  So it will do denormalization for the same of a given batch query, but it will not cache for further batch queries.  This usually comes up when the same query is needed at multiple levels in the query.

dylanahsmith:
Should be fixed by https://github.com/Shopify/graphql-batch/pull/53 now

dylanahsmith:
Reopening this because #53 was reverted due to it resulting in loader re-use across mutation fields

PR #56, opened 2017-06-06
Prevent loader re-use after mutations

dylanahsmith:
cc @staugaard

## Problem

The loader re-use added by https://github.com/Shopify/graphql-batch/pull/53 also re-used the loaders and their resolved promises after a mutation field is resolved.

[GraphQL::GraphQLTest#test_mutation_execution](https://github.com/Shopify/graphql-batch/blob/v0.3.3/test/graphql_test.rb#L297) was a test that was supposed to catch this bug, but it was using `CounterLoader.for(ctx).load` where load was defined in CounterLoader with `def load(key=Object.new)`, so it was actually using a unique key for each load which prevented it from ever re-using the result of a previous load.

## Solution

I've reverted https://github.com/Shopify/graphql-batch/pull/53 to fix the bug and updated the test to use override `cache_key` to use the `object_id` of the array that is used to hold the counter.

If the test is run without the revert, then it has the expected failures of

```
  1) Failure:
GraphQL::GraphQLTest#test_mutation_execution [test/graphql_test.rb:315]:
--- expected
+++ actual
@@ -1 +1 @@
-{"data"=>{"count1"=>0, "incr1"=>{"value"=>1, "load_value"=>1}, "count2"=>1, "incr2"=>{"value"=>2, "load_value"=>2}}}
+{"data"=>{"count1"=>0, "incr1"=>{"value"=>0, "load_value"=>0}, "count2"=>0, "incr2"=>{"value"=>0, "load_value"=>0}}}
```

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -44,8 +44,8 @@ module GraphQL::Batch
     end
 
     def resolve #:nodoc:
+      return if resolved?
       load_keys = queue
-      return if load_keys.empty?
       @queue = nil
       perform(load_keys)
       check_for_broken_promises(load_keys)
@@ -63,6 +63,10 @@ module GraphQL::Batch
       else
         resolve
       end
+    end
+
+    def resolved?
+      @queue.nil? || @queue.empty?
     end
 
     protected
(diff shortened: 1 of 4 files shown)


dylanahsmith:
I will need to reopen https://github.com/Shopify/graphql-batch/issues/43 after this

staugaard:
Maybe we could clear the loader cache after executing a mutation

PR #57, opened 2017-06-06
feat(GraphQL::Batch::SetupMultiplex) use multiplex instrumentation to share a batching context

rmosolgo:
This patch adds support for a shared batching context when multiple queries are run concurrently. Is it something you're interested in for this gem?

Single queries are still supported because graphql 1.6+ treats `Schema#execute` as a special case of `Schema#multiplex`: a multiplex with one query. So it's still covered by `instrument(:multiplex,...)`.

### DIFF
--- a/test/multiplex_test.rb
+++ b/test/multiplex_test.rb
@@ -1,0 +1,51 @@
+require_relative 'test_helper'
+
+class GraphQL::MultiplexTest < Minitest::Test
+  attr_reader :queries
+
+  def setup
+    @queries = []
+    QueryNotifier.subscriber = ->(query) { @queries << query }
+  end
+
+  def teardown
+    QueryNotifier.subscriber = nil
+  end
+
+  def schema_multiplex(*args)
+    ::Schema.multiplex(*args)
+  end
+
+
+  def test_batched_find_by_id
+    query_string = <<-GRAPHQL
+      query FetchTwoProducts($id1: ID!, $id2: ID!){
+        first: product(id: $id1) { id, title }
+        second: product(id: $id2) { id, title }
+      }
+    GRAPHQL
+
+    results = schema_multiplex([
+      { query: query_string, variables: { "id1" => "1", "id2" => "2"} },
+      { query: query_string, variables: { "id1" => "1", "id2" => "3"}},
+    ])
+
+    expected_1 = {
+      "data" => {
+        "first" => { "id" => "1", "title" => "Shirt" },
+        "second" => { "id" => "2", "title" => "Pants" },
+      }
+    }
+
+    expected_2 = {
+      "data" => {
+        "first" => { "id" => "1", "title" => "Shirt" },
+        "second" => { "id" => "3", "title" => "Sweater" },
+      }
+    }
+
+    assert_equal expected_1, results.first
+    assert_equal expected_2, results.last
+    assert_equal ["Product/1,2,3"], queries
+  end
+end
--- a/test/multiplex_test.rb
+++ b/test/multiplex_test.rb
@@ -12,11 +12,6 @@ class GraphQL::MultiplexTest < Minitest::Test
     QueryNotifier.subscriber = nil
   end
 
-  def schema_multiplex(*args)
-    ::Schema.multiplex(*args)
-  end
-
-
   def test_batched_find_by_id
     query_string = <<-GRAPHQL
       query FetchTwoProducts($id1: ID!, $id2: ID!){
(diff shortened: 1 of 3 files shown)


test/multiplex_test.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def test_batched_find_by_id
+    query_string = <<-GRAPHQL
+      query FetchTwoProducts($id1: ID!, $id2: ID!){
+        first: product(id: $id1) { id, title }
+        second: product(id: $id2) { id, title }
+      }
+    GRAPHQL
+
+    results = schema_multiplex([
```
dylanahsmith: You can just use `Schema.multiplex` directly.  The reason test/graphql_test.rb has a schema_execute method is that execution_strategy_test.rb overrides it to do the same tests.

test/multiplex_test.rb
```diff
@@ -0,0 +1,51 @@
...
+      "data" => {
+        "first" => { "id" => "1", "title" => "Shirt" },
+        "second" => { "id" => "3", "title" => "Sweater" },
+      }
+    }
+
+    assert_equal expected_1, results.first
+    assert_equal expected_2, results.last
+    assert_equal ["Product/1,2,3"], queries
+  end
```
dylanahsmith: I tried making the following changes to the test

```diff
diff --git a/test/multiplex_test.rb b/test/multiplex_test.rb
index c120614..6748aab 100644
--- a/test/multiplex_test.rb
+++ b/test/multiplex_test.rb
@@ -24,9 +24,15 @@ class GraphQL::MultiplexTest < Minitest::Test
         second: product(id: $id2) { id, title }
       }
     GRAPHQL
+    mutation_string = <<-GRAPHQL
+      mutation {
+        counter_loader
+      }
+    GRAPHQL
 
     results = schema_multiplex([
       { query: query_string, variables: { "id1" => "1", "id2" => "2"} },
+      { query: mutation_string, context: { counter: 0 } },
       { query: query_string, variables: { "id1" => "1", "id2" => "3"}},
     ])
 
@@ -39,13 +45,20 @@ class GraphQL::MultiplexTest < Minitest::Test
 
     expected_2 = {
       "data" => {
+        "counter_loader" => 0,
+      }
+    }
+
+    expected_3 = {
+      "data" => {
         "first" => { "id" => "1", "title" => "Shirt" },
         "second" => { "id" => "3", "title" => "Sweater" },
       }
     }
 
     assert_equal expected_1, results.first
-    assert_equal expected_2, results.last
+    assert_equal expected_2, results[1]
+    assert_equal expected_3, results.last
     assert_equal ["Product/1,2,3"], queries
   end
 end
```

and it still passed, even though there is a mutation in between the queries.

What should be the expectations around the ordering of the operations?  The [graphql-ruby multiplex documentation](http://graphql-ruby.org/queries/multiplex.html) says it will "run queries concurrently", but also says this feature is for [Apollo Client’s query batching](http://dev.apollodata.com/core/network.html#query-batching) which "is a transport-level mechanism" where I would expect the order of operations would be preserved.

rmosolgo: That's a great point, and it's something I overlooked. It makes no special treatment for mutations, but it could. (Apparently the JS server doesn't make special treatment either: https://github.com/apollographql/graphql-server/issues/176#issuecomment-290956816)

Maybe we could do a flow like this: 

- Shift the next operation from the set
  - If it's a mutation:
    - Run any already-gathered queries concurrently
    - Run it in isolation 
  - If it's a query, add it to the next multiplex
- When the set is empty: 
  - Run any already-gathered queries concurrently 

I think that would maintain the expected ordering for mutations: queries before mutations would not reflect their changes, but queries after mutations _would_ reflect mutations. 

However, now that I write it, I don't know if the incoming queries _are_ ordered. Maybe they come as an unordered set. 

If that's the case, maybe a better flow is mutate-then-read: 

- For each operation in the set, if it's a mutation, run its root fields
- Then, fullfill all other fields concurrently. 

That way, all reads in the multiplex reflect the changes of all mutations in the multiplex. 

Hmm, what if the mutations are order-dependent? :head_explodes:

dylanahsmith: Interesting, so it looks like they intentionally have the queries running in parallel according to the apollo server implementation.  Given that, and the fact that graphql-ruby already follows that approach and documents it more clearly, I guess there is no need to be concerned about it here.

I suppose a client or a server could always choose to re-order the operations and multiplex them based on their desired semantics.

> Hmm, what if the mutations are order-dependent? :head_explodes:

In that case I would put all the mutations in the same mutation query as separate field selections, although that is easier when using dynamic queries.

test/multiplex_test.rb
dylanahsmith: I'm OK with merging this PR after this nitpick is addressed.

etripier:
This is awesome! It fixes all the issues I was having integrating `graphql-batch` with `react-apollo`. Just chiming in to show interest in getting this merged. Thanks @rmosolgo!

test/multiplex_test.rb
rmosolgo: 🤦‍♂️  Sorry I forgot about this feedback all this time! I just pushed a fix for it.

PR #58, opened 2017-07-12
Deprecated ExecutionStrategy has better query performance

rf-:
Hi! This PR isn't really meant to be merged, just to demonstrate an issue I ran into. It's probably more of an issue with GraphQL than with GraphQL::Batch, but I figured I'd start by opening it here for discussion.

I'm not 100% sure what's going on, but the deprecated GraphQL::Batch::ExecutionStrategy class seems to do a better job of aggregating queries than the non-deprecated integration. Specifically, the test case in this PR passes with `LegacySchema` but fails with `Schema`:

      1) Failure:
    GraphQL::GraphQLTest#test_deeply_nested_lazy_queries [/Users/rfitz/src/vendor/graphql-batch/test/graphql_test.rb:320]:
    --- expected
    +++ actual
    @@ -1 +1 @@
    -["Product?limit=2", "Product/1,2/variants", "ProductVariant/1,2,4,5,6/images"]
    +["Product?limit=2", "Product/1,2/variants", "ProductVariant/1,2/images", "ProductVariant/4,5,6/images"]

The difference between this test and the existing `test_load_list_of_objects_with_loaded_field` is just an extra layer of laziness in the query.

I'm happy to just use ExecutionStrategy for now, but what's the path forward here? Should I open an issue on the GraphQL gem, or is this something that can be addressed at the GraphQL::Batch layer?

### DIFF
--- a/test/graphql_test.rb
+++ b/test/graphql_test.rb
@@ -279,6 +279,47 @@ class GraphQL::GraphQLTest < Minitest::Test
     assert_equal ["Product?limit=2", "Product/1,2/variants", "ProductVariant/1,2,4,5,6/images"], queries
   end
 
+  def test_deeply_nested_lazy_queries
+    query_string = <<-GRAPHQL
+      {
+        products(first: 2) {
+          this_product {
+            variants {
+              id
+              image_ids
+            }
+          }
+        }
+      }
+    GRAPHQL
+    result = schema_execute(query_string)
+    expected = {
+      "data" => {
+        "products" => [
+          {
+            "this_product" => {
+              "variants" => [
+                { "id" => "1", "image_ids" => ["4"] },
+                { "id" => "2", "image_ids" => ["5"] },
+              ],
+            },
+          },
+          {
+            "this_product" => {
+              "variants" => [
+                { "id" => "4", "image_ids" => [] },
+                { "id" => "5", "image_ids" => [] },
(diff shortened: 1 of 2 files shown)


rf-:
BTW, if I add more layers of `this_product { ... }` to the query, the performance in GraphQLTest continues to get worse while ExecutionStrategyTest stays the same.

rmosolgo:
Thanks for the failing test case here! I "borrowed" graphql-batch's breadth-first batching and added it to graphql-ruby internals. 

I think that's where the real bug is here: my port doesn't have perfect parity! 

I'll track down this issue in graphql-ruby and get back to you. Thanks again for the failing test -- hopefully we can make it pass :)

dylanahsmith:
Yes, this does look like a bug.

I tested with the first version of graphql-ruby that added lazy_resolve and the first version of graphql-batch that added support for it and the bug was still present, so it doesn't seem like a regression in the lazy_resolve support.

I haven't dug into why this is happening yet

rmosolgo:
Pushed a fix in graphql-ruby here: https://github.com/rmosolgo/graphql-ruby/pull/835

I'll release 1.6.5 soon with this patch, then this test will pass! (It passes for me locally.)

rf-:
Thanks for the quick fix @rmosolgo! It fixes both the test in this PR and the version with more layers of nesting added. 💸

PR #61, opened 2017-08-08
fix: Promise namespacing inconsitency

### LINKED ISSUE #60: Promise Namespacing Issue
There is a Namespacing inconsistency in `GraphQL::Batch::Loader` between the Promises returned by `load` and `load_many` causing them to return different classes.

```ruby
   def load(key)
      cache[cache_key(key)] ||= begin
        queue << key
        Promise.new.tap { |promise| promise.source = self } ### returns GraphQL::Batch::Promise
      end
    end

    def load_many(keys)
      ::Promise.all(keys.map { |key| load(key) })  ### returns Promise 
    end
```

Is there a particular reason for this? I can't seem to reason one.

jpserra:
Fixes #60

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -40,7 +40,7 @@ module GraphQL::Batch
     end
 
     def load_many(keys)
-      ::Promise.all(keys.map { |key| load(key) })
+      Promise.all(keys.map { |key| load(key) })
     end
 
     def resolve #:nodoc:


dylanahsmith:
If we are going to address #60, then I would prefer to do the refactoring necessary to get rid of GraphQL::Batch::Promise entirely.  Initially it was added to override `Promise#wait` but now `Promise#source=` is used to avoid the need for a derived class.

PR #64, opened 2017-08-11
Add Active Record AssociationLoader and RecordLoader examples

### LINKED ISSUE #24: Document how one could preload has_many associations

dylanahsmith:
Fixes #24

This adds a couple of examples for a RecordLoader and AssociationLoader that work with Active Record and links to those examples in the README.

Each of these handle a problem that a naive loader would have:
* RecordLoader type casts keys to avoid to fulfilling the wrong key when the wrong type is provided
* AssociationLoader preloads the association on all records, even if some share the same id

These are just examples, since I didn't want to add an active record dependency to this library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,12 @@ The loader class can be used from the resolve proc for a graphql field by callin
 resolve -> (obj, args, context) { RecordLoader.for(Product).load(args["id"]) }
 ```
 
+Although this library doesn't have a dependency on active record,
+the [examples directory](examples) has record and association loaders
+for active record which handles edge cases like type casting ids
+and overriding GraphQL::Batch::Loader#cache_key to load associations
+on records with the same id.
+
 ### Promises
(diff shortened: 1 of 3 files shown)


xuorig:
Cool, should help newcomers a lot.

I wrote this a while ago for an easy plug and play solution in graphql ruby: https://github.com/xuorig/graphql-active_record_batcher

might be cool to link it 🤷‍♂️

dylanahsmith:
> I'll PR to add a link to the graphql-ruby integration in a follow up PR?

Yes, please.  If you add a RecordLoader to your repo, then we could just link directly to the loaders in that repo and won't need examples in this repo.

PR #65, opened 2017-08-31
Return a normal Promise object from GraphQL::Batch::Loader#load

### LINKED ISSUE #60: Promise Namespacing Issue

dylanahsmith:
Fixes #60 (cc @jpserra)

@arthurschreiber would this remove the need for https://github.com/lgierth/promise.rb/pull/35 ?

As I mentioned in #60 

> The reason for the difference is more of an implementation detail. The only difference between ::Promise and GraphQL::Batch::Promise is that GraphQL::Batch::Promise overrides defer so that we can use GraphQL::Batch::Executor.current.loading in tests to detect when an unbatched query is performed in a test.

so I refactored the code so that GraphQL::Batch::Loader's fulfill and reject methods call `executor.around_promise_callbacks do` around the actually fulfillment or rejection of the promise, allowing the executor to set its loading flag to `false` before promise callbacks are called and reset back to `true` for the remainder of the loaders perform method.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -12,10 +12,7 @@ module GraphQL::Batch
           " `GraphQL::Batch::Setup` as a query instrumenter if using with `graphql-ruby`"
       end
 
-      executor.loaders[loader_key] ||= new(*group_args).tap do |loader|
-        loader.loader_key = loader_key
-        loader.executor = executor
-      end
+      executor.loader(loader_key) { new(*group_args) }
     end
 
     def self.loader_key_for(*group_args)
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -32,7 +32,7 @@ module GraphQL::Batch
     def load(key)
       cache[cache_key(key)] ||= begin
         queue << key
-        Promise.new.tap { |promise| promise.source = self }
+        ::Promise.new.tap { |promise| promise.source = self }
       end
     end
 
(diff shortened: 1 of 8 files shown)


lib/graphql/batch/loader.rb
```diff
@@ -92,6 +95,14 @@ def cache_key(load_key)
 
     private
 
+    def finish_resolve(key)
+      promise = promise_for(key)
+      return yield(promise) unless executor
```
xuorig: is this line necessary since it will raise if not wrapped in `GraphQL::Batch.batch` in `for`?

dylanahsmith: The executor is mostly just used for storing loaders so they can be looked up for grouping, however, an alternative would be to store loaders in a context.  We started supporting using loaders without an executor in https://github.com/Shopify/graphql-batch/pull/35.

xuorig: oh yes, of course! 👍

PR #68, opened 2017-11-06
Allow custom executors

chrisbutcher:
* Changed: GraphQL::Batch.use can take optional `custom_executor`
  * `GraphQL::Batch::Setup` and `GraphQL::Batch::SetupMultiplex` updated to hold reference to the executor class (defaults to GraphQL::Batch::Executor) as instance variables.

@dylanahsmith @eapache @swalkinshaw 👋 ~~I'm looking to verify my approach here before continuing. See the linked Shopify/shopify PR for more context.~~ Updated according PR review comments so far.

### DIFF
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -19,14 +19,14 @@ module GraphQL
       end
     end
 
-    def self.use(schema_defn)
+    def self.use(schema_defn, executor_class: GraphQL::Batch::Executor)
       schema = schema_defn.target
       if GraphQL::VERSION >= "1.6.0"
-        instrumentation = GraphQL::Batch::SetupMultiplex.new(schema)
+        instrumentation = GraphQL::Batch::SetupMultiplex.new(schema, executor_class: executor_class)
         schema_defn.instrument(:multiplex, instrumentation)
         schema_defn.instrument(:field, instrumentation)
       else
-        instrumentation = GraphQL::Batch::Setup.new(schema)
+        instrumentation = GraphQL::Batch::Setup.new(schema, executor_class: executor_class)
         schema_defn.instrument(:query, instrumentation)
         schema_defn.instrument(:field, instrumentation)
       end
--- a/lib/graphql/batch/setup_multiplex.rb
+++ b/lib/graphql/batch/setup_multiplex.rb
@@ -1,11 +1,12 @@
 module GraphQL::Batch
   class SetupMultiplex
-    def initialize(schema)
+    def initialize(schema, executor_class:)
       @schema = schema
+      @executor_class = executor_class
     end
 
     def before_multiplex(multiplex)
-      Setup.start_batching
+      Setup.start_batching(@executor_class)
     end
 
     def after_multiplex(multiplex)
--- a/lib/graphql/batch/setup.rb
+++ b/lib/graphql/batch/setup.rb
@@ -1,9 +1,9 @@
 module GraphQL::Batch
   class Setup
     class << self
-      def start_batching
+      def start_batching(executor_class)
         raise NestedError if GraphQL::Batch::Executor.current
-        GraphQL::Batch::Executor.current = GraphQL::Batch::Executor.new
+        GraphQL::Batch::Executor.current = executor_class.new
       end
 
       def end_batching
@@ -27,7 +27,7 @@ module GraphQL::Batch
 
       def before_query(query)
         warn "Deprecated graphql-batch setup `instrument(:query, GraphQL::Batch::Setup)`, replace with `use GraphQL::Batch`"
-        start_batching
+        start_batching(GraphQL::Batch::Executor)
       end
 
       def after_query(query)
@@ -35,12 +35,13 @@ module GraphQL::Batch
       end
     end
 
-    def initialize(schema)
+    def initialize(schema, executor_class:)
       @schema = schema
+      @executor_class = executor_class
     end
 
     def before_query(query)
-      Setup.start_batching
+      Setup.start_batching(@executor_class)
     end
 
     def after_query(query)
--- a/test/custom_executor_test.rb
+++ b/test/custom_executor_test.rb
@@ -1,0 +1,31 @@
+require_relative 'test_helper'
+
+class GraphQL::Batch::CustomExecutorTest < Minitest::Test
+  class MyCustomExecutor < GraphQL::Batch::Executor
+    @@call_count = 0
+
+    def self.call_count
+      @@call_count
+    end
+
+    def around_promise_callbacks
+      @@call_count += 1
+
+      super
+    end
+  end
+
+  def test_custom_executor_class
+    schema = GraphQL::Schema.define do
+      query ::QueryType
+      mutation ::MutationType
+
+      use GraphQL::Batch, executor_class: MyCustomExecutor
+    end
+
+    query_string = '{ product(id: "1") { id } }'
+    schema.execute(query_string)
+
+    assert MyCustomExecutor.call_count > 0
+  end
+end


lib/graphql/batch.rb
```diff
@@ -12,20 +12,22 @@ class NestedError < StandardError; end
...
+    def self.use(schema_defn, executor_class: GraphQL::Batch::Executor)
       schema = schema_defn.target
       if GraphQL::VERSION >= "1.6.0"
+        GraphQL::Batch::SetupMultiplex.executor_class = executor_class
         instrumentation = GraphQL::Batch::SetupMultiplex.new(schema)
         schema_defn.instrument(:multiplex, instrumentation)
         schema_defn.instrument(:field, instrumentation)
       else
+        GraphQL::Batch::Setup.executor_class = executor_class
         instrumentation = GraphQL::Batch::Setup.new(schema)
```
dylanahsmith: Store the state on the instance rather than the class.  It is only supposed to affect the schema it is called on.

lib/graphql/batch.rb
```diff
@@ -12,20 +12,22 @@ class NestedError < StandardError; end
     def self.batch
       raise NestedError if GraphQL::Batch::Executor.current
       begin
-        GraphQL::Batch::Executor.current = GraphQL::Batch::Executor.new
+        GraphQL::Batch::Executor.current = GraphQL::Batch::Executor.new # TODO
```
dylanahsmith: Changing this isn't important for our purposes, so you could just remove the `# TODO` comment.

chrisbutcher:
Thanks @dylanahsmith ! I've updated the PR in response to your suggestions. Any suggestions on how/if we should unit test this change? We don't have a SetupTest, nor coverage on `GraphQL::Batch.use`.

The only testable changes appear to be in `GraphQL::Batch.use`'s forwarding of `executor_class`, and the expectation in `start_batching` that the passed-in Executor class instantiated.

dylanahsmith:
> Any suggestions on how/if we should unit test this change? We don't have a SetupTest, nor coverage on GraphQL::Batch.use.

Integration style tests would be preferred for something like that, since it is mean to be used through graphql-ruby rather than directly. We test `GraphQL::Batch.use` through [`use GraphQL::Batch`](https://github.com/Shopify/graphql-batch/blob/v0.3.5/test/support/schema.rb#L162) right now

test/graphql_test.rb
```diff
@@ -374,4 +374,35 @@ def test_mutation_batch_subselection_execution
...
+      use GraphQL::Batch, executor_class: MyCustomExecutor
+    end
+
+    custom_executor = MyCustomExecutor.new
+    MyCustomExecutor.stubs(:new).returns(custom_executor)
+
+    query_string = '{ product(id: "1") { id } }'
+    schema.execute(query_string)
+
+    refute custom_executor.loaders_executed.empty?
```
chrisbutcher: Possibly related to https://github.com/Shopify/shopify/pull/136940#issuecomment-342542248?
I noticed that this ended up containing two references to the same instance of RecordLoader.

chrisbutcher:
Updated. I ended up adding a gem dependency (mocha) to make assertions on some trivial behaviour defined on a custom executor.

test/graphql_test.rb
```diff
@@ -374,4 +374,35 @@ def test_mutation_batch_subselection_execution
...
+  def test_custom_executor_class
+    schema = GraphQL::Schema.define do
+      query ::QueryType
+      mutation ::MutationType
+
+      use GraphQL::Batch, executor_class: MyCustomExecutor
+    end
+
+    custom_executor = MyCustomExecutor.new
+    MyCustomExecutor.stubs(:new).returns(custom_executor)
```
dylanahsmith: I would prefer if you at least use a class variable for the loaders_executed array so that you don't need to mess with the implementation using stubbing.

test/graphql_test.rb
```diff
@@ -374,4 +374,35 @@ def test_mutation_batch_subselection_execution
...
+      super
+    end
+
+    def around_promise_callbacks(loader)
+      @loaders_executed << loader
+      super
+    end
+  end
+
+  def test_custom_executor_class
```
dylanahsmith: I think this should move to another test class, since this one gets re-used in GraphQL::ExecutionStrategyTest which tests the deprecated code that we don't support custom executors for.

chrisbutcher:
@swalkinshaw I'm looking into this now. It seems to be because we call the instrumented method twice via `finish_resolve` https://github.com/Shopify/graphql-batch/blob/master/lib/graphql/batch/loader.rb#L101, because we call `finish_resolve` in both `fulfill` and `reject`: https://github.com/Shopify/graphql-batch/blob/master/lib/graphql/batch/loader.rb#L69-L79

It does this regardless of whether an existing collection id is queried in the query mentioned [here](https://github.com/Shopify/shopify/pull/136940#issuecomment-342542248).

test/graphql_test.rb
```diff
@@ -374,4 +374,35 @@ def test_mutation_batch_subselection_execution
...
+  class MyCustomExecutor < GraphQL::Batch::Executor
+    attr_reader :loaders_executed
+
+    def initialize
+      @loaders_executed = []
+      super
+    end
+
+    def around_promise_callbacks(loader)
+      @loaders_executed << loader
```
dylanahsmith: This is called each time a promise is fulfilled, not for each loader being executed.  E.g. if 4 keys are being loaded by a loader, then this will be called for all 4 keys rather than the one load.

chrisbutcher: > This is called each time a promise is fulfilled, not for each loader being executed.

Thanks for the clarification!

It looks like `around_promise_callbacks` is also invoked via `Loader#reject`. So that's twice per field, even if the field is fulfilled. `reject` is called even when the loader finds a result, via `check_for_broken_promises`. This explains the double logging in my current implementation.

```
lib/graphql/batch/loader.rb:78:in `reject'
lib/graphql/batch/loader.rb:131:in `block in check_for_broken_promises'
lib/graphql/batch/loader.rb:130:in `each'
lib/graphql/batch/loader.rb:130:in `check_for_broken_promises'
lib/graphql/batch/loader.rb:48:in `resolve'
lib/graphql/batch/executor.rb:34:in `resolve'
lib/graphql/batch/loader.rb:56:in `wait'
lib/promise.rb:129:in `wait'
lib/promise.rb:129:in `wait'
lib/promise.rb:129:in `wait'
lib/promise.rb:80:in `sync'
lib/graphql/field.rb:309:in `public_send'
```

chrisbutcher: I was able to  around the double-logging by resetting `@start_time` to nil in `around_promise_callbacks`, like so:

```ruby
module GraphModel
  class TimedBatchExecutor < GraphQL::Batch::Executor
 
    # etc.
    
    def resolve(loader)
      @start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)

      super
    end

    def around_promise_callbacks(loader)
      return super unless @start_time

      end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)

      TIMINGS[loader.loader_key] ||= []
      TIMINGS[loader.loader_key] << end_time - @start_time

      @start_time = nil

      super
    end
  end
end
```

This prevents the second call to `around_promise_callbacks` (via Executor#resolve -> Loader#resolve, #check_for_broken_promises, #reject, #finish_resolve) from logging the time between the original `@start_time` and the time it runs again.

chrisbutcher:
Another follow-up change after some discussion with @swalkinshaw.

I think we can get around modifying `around_promise_callbacks`'s signature to take a `loader` by simply keeping a reference to the loader we already receive in Executor#resolve. Will push up commit shortly.

chrisbutcher:
Updated and ready for another review

lib/graphql/batch/setup_multiplex.rb
```diff
@@ -1,11 +1,12 @@
 module GraphQL::Batch
   class SetupMultiplex
-    def initialize(schema)
+    def initialize(schema, executor_class: executor_class)
```
swalkinshaw: 🚨  hold up.

> lib/graphql/batch/setup_multiplex.rb:3: warning: circular argument reference - executor_class

chrisbutcher: 👀

chrisbutcher:
```diff
- def initialize(schema, executor_class: executor_class)
+ def initialize(schema, executor_class:) 
```
✅

Issue #70, opened 2017-11-13
NoMethodError - undefined method `clear' for nil:NilClass


emwalker:
I'm seeing this stack trace when I use graphql (`1.7.5`) and graphql-batch (`0.3.8`) together with subscriptions:
```
NoMethodError - undefined method `clear' for nil:NilClass:
  graphql-batch (0.3.8) lib/graphql/batch/setup.rb:22:in `ensure in block (2 levels) in
  graphql-batch (0.3.8) lib/graphql/batch/setup.rb:22:in `block (2 levels) in
  graphql (1.7.5) lib/graphql/relay/mutation/resolve.rb:18:in `call'
  graphql (1.7.5) lib/graphql/field.rb:230:in `resolve'
  graphql (1.7.5) lib/graphql/execution/execute.rb:254:in `call'
  graphql (1.7.5) lib/graphql/schema/middleware_chain.rb:45:in `invoke_core'
  graphql (1.7.5) lib/graphql/schema/middleware_chain.rb:38:in `invoke'
  graphql (1.7.5) lib/graphql/execution/execute.rb:108:in `resolve_field'
  graphql (1.7.5) lib/graphql/execution/execute.rb:72:in `block (2 levels) in
  graphql (1.7.5) lib/graphql/tracing.rb:56:in `block in trace'
  graphql (1.7.5) lib/graphql/tracing.rb:70:in `call_tracers'
  graphql (1.7.5) lib/graphql/tracing.rb:56:in `trace'
  graphql (1.7.5) lib/graphql/execution/execute.rb:71:in `block in resolve_selection'
  graphql (1.7.5) lib/graphql/execution/execute.rb:64:in `resolve_selection'
  graphql (1.7.5) lib/graphql/execution/execute.rb:36:in `block in resolve_root_selection'
  graphql (1.7.5) lib/graphql/tracing.rb:56:in `block in trace'
  graphql (1.7.5) lib/graphql/tracing.rb:70:in `call_tracers'
  graphql (1.7.5) lib/graphql/tracing.rb:56:in `trace'
  graphql (1.7.5) lib/graphql/execution/execute.rb:32:in `resolve_root_selection'
  graphql (1.7.5) lib/graphql/execution/multiplex.rb:108:in `begin_query'
  graphql (1.7.5) lib/graphql/execution/multiplex.rb:81:in `block in run_as_multiplex'
[...]

PR #72, opened 2017-11-29
Allow nested queries.

emwalker:
This PR removes `GraphQL::Batch:: NestedError` and the associated restriction on nested queries. See https://github.com/Shopify/graphql-batch/issues/70 for details.

### DIFF
--- a/test/batch_test.rb
+++ b/test/batch_test.rb
@@ -9,10 +9,19 @@ class GraphQL::BatchTest < Minitest::Test
   end
 
   def test_nested_batch
-    GraphQL::Batch.batch do
-      assert_raises(GraphQL::Batch::NestedError) do
-        GraphQL::Batch.batch {}
+    promise1 = nil
+    promise2 = nil
+
+    product = GraphQL::Batch.batch do
+      promise1 = RecordLoader.for(Product).load(1)
+      GraphQL::Batch.batch do
+        promise2 = RecordLoader.for(Product).load(1)
       end
+      promise1
     end
+
+    assert_equal 'Shirt', product.title
+    assert_equal promise1, promise2
+    assert_nil GraphQL::Batch::Executor.current
   end
 end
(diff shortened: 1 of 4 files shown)


test/batch_test.rb
```diff
@@ -9,10 +9,13 @@ def test_batch
   end
 
   def test_nested_batch
-    GraphQL::Batch.batch do
-      assert_raises(GraphQL::Batch::NestedError) do
-        GraphQL::Batch.batch {}
+    product = GraphQL::Batch.batch do
+      promise = RecordLoader.for(Product).load(1)
+      GraphQL::Batch.batch do
+        RecordLoader.for(Product).load(1)
```
dylanahsmith: You aren't testing anything with this.  Perhaps you could assert that it returns the same promise object as the one above.

test/batch_test.rb
```diff
@@ -9,10 +9,13 @@ def test_batch
...
       end
+      promise
```
dylanahsmith: You should also test a load here to make sure the executor is still setup.  Again, this could just be asserting that `RecordLoader.for(Product).load(1)` returns the same promise.

test/batch_test.rb
```diff
@@ -9,10 +9,13 @@ def test_batch
...
     end
+    assert_equal 'Shirt', product.title
```
dylanahsmith: Maybe also a `assert_nil GraphQL::Batch::Executor.current` check to make sure the cleanup code works on the outermost `GraphQL::Batch.batch`

test/batch_test.rb
emwalker: Done.

test/batch_test.rb
emwalker: Done.

test/batch_test.rb
emwalker: Done.

emwalker:
@dylanahsmith how does the revised test look?

emwalker:
@dylanahsmith what are the chances of getting a patch version rev to `v0.3.9` and publishing a new version of the gem?  (I should have included one in my PR.)

dylanahsmith:
Worry for the delay in my response, I just got back from vacation.  Just released v0.3.9

PR #74, opened 2017-12-13
Fix NoExecutorError

xuorig:
When ending a batch with no executor, we would try to raise a `NoExecutorError` in `Executor` but a name error would occur because that class is defined in `Loader`.

In this PR I'm moving this error to `GraphQL::Batch` instead since it may happen both in `Loader` and `Executor`.

Also added a test to ensure this code path is tested.

### DIFF
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -3,12 +3,11 @@ if Gem::Version.new(GraphQL::VERSION) < Gem::Version.new("1.3")
   warn "graphql gem versions less than 1.3 are deprecated for use with graphql-batch, upgrade so lazy_resolve can be used"
 end
 require "promise.rb"
-require_relative "batch/loader"
 
 module GraphQL
   module Batch
     BrokenPromiseError = ::Promise::BrokenError
-    NoExecutorError = Loader::NoExecutorError
+    class NoExecutorError < StandardError; end
 
     def self.batch
       begin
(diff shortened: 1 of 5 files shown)


lib/graphql/batch/executor.rb
```diff
@@ -20,7 +20,8 @@ def start_batch(executor_class)
       def end_batch
         executor = current
         unless executor
-          raise NoExecutorError, 'Cannot end a batch without an Executor.'
+          raise GraphQL::Batch::NoExecutorError,
```
dylanahsmith: With NoExecutorError defined on GraphQL::Batch there is no need to reference it directly, just using `NoExecutorError` as was done before should be sufficient.

lib/graphql/batch.rb
```diff
@@ -7,6 +7,7 @@
 module GraphQL
   module Batch
     BrokenPromiseError = ::Promise::BrokenError
+    class NoExecutorError < StandardError; end
```
dylanahsmith: You have defined a second exception here without removing GraphQL::Batch::Loader::NoExecutorError.  You should instead make this an alias with `NoExecutorError = Loader::NoExecutorError` and use `deprecate_constant :NoExecutorError` in GraphQL::Batch::Loader.

xuorig: Ah good catch, done.

lib/graphql/batch.rb
```diff
@@ -3,10 +3,12 @@
   warn "graphql gem versions less than 1.3 are deprecated for use with graphql-batch, upgrade so lazy_resolve can be used"
 end
 require "promise.rb"
+require_relative "batch/loader"
```
xuorig: moved this one up because we refer to it in the constants below

PR #75, opened 2017-12-13
Allow a custom executor class in `GraphQL::Batch.batch`

xuorig:
Follow up to https://github.com/Shopify/graphql-batch/pull/68

Would be great to be able to use this using the `GraphQL::Batch.batch` API. WDYT?

cc @chrisbutcher


xuorig:
Policial doesnt seem to know `positive?` is not a thing in older rubies 🤷‍♂️

dylanahsmith:
I don't think Policial should even be checking this repo, since there is no rubocop config file, in which I would otherwise suggest that we just ignore the cop.  I've changed the Policial settings for the repo to not check ruby files.

test/batch_test.rb
```diff
@@ -8,6 +8,29 @@ def test_batch
     assert_equal 'Shirt', product.title
   end
 
+  class MyCustomExecutor < GraphQL::Batch::Executor
```
dylanahsmith: I would prefer if you moved the new test to GraphQL::Batch::CustomExecutorTest so that we can use the MyCustomExecutor in there rather than copying it here.

xuorig: 👍 moved it there

PR #80, opened 2018-03-06
Do not try to reject pending promises.

arthurschreiber:
This fixes an issue where `Loader#check_for_broken_promises` was generating a large amount of unused `Promise::BrokenError` objects and strings in case there were no broken promises after the loader finished execution.

By skipping the error object creation for promises that are no longer pending, we can also skip generating these objects.


dylanahsmith:
Thanks for the fix

joshjcarrier:
@dylanahsmith can we create a release that includes this please?

dylanahsmith:
I've made a new release that includes this

Issue #82, opened 2018-04-10
Ask for Sequel users


razertory:
Could there be a [Sequel](https://github.com/jeremyevans/sequel) `association_loader.rb` example , just like the Active Record one?

dylanahsmith:
We don't plan on writing an example for Sequel.  Is that what you are offering?

How different would it be from the current association_loader.rb?  Wouldn't it just be replacing the last three methods https://github.com/Shopify/graphql-batch/blob/4ae285e4c2b873a5338ae555abeb5bc58647ec46/examples/association_loader.rb#L37-L47 with something non-active record specific?

razertory:
Sequel's association_loader example is just here and thanks to @rmosolgo , his example woks.
```ruby
class ForeignKeyLoader < GraphQL::Batch::Loader
  def initialize(model, foreign_key)
    @model = model
    @foreign_key = foreign_key
  end

  def perform(foreign_value_sets)
    foreign_values = foreign_value_sets.flatten.uniq
    records = @model.where(@foreign_key => foreign_values).to_a

    foreign_value_sets.each do |foreign_value_set|
      matching_records = records.select { |r| foreign_value_set.include?(r.send(@foreign_key)) }
      fulfill(foreign_value_set, matching_records)
    end
  end
end
```

dylanahsmith:
That isn't loading Sequel's associations.

I could see that a ForeignKeyLoader might be the next best thing and could be implemented in a way that works with active record and Sequel.

dylanahsmith:
It looks like the RecordLoader can already act as a foreign key loader for a `has_one` association (e.g. `RecordLoader.for(ProductImage, column: :product_id).load(product_id)`).  It just needs to be extended to allow multiple records by a foreign key.  This could be done by adding a `unique: false` parameter.

Thoughts on changing the RecordLoader example as shown in this [foreign-key-loader-example branch](https://github.com/Shopify/graphql-batch/compare/foreign-key-loader-example)?

Likely the column type casting would still need to be removed to work with Sequel.

ayrton:
@jeremyevans any pointers as to how the `AssociationLoader` would look like for Sequel?

jeremyevans:
Unfortunately, I do not have the time to research the code and come up with a tested example.  In terms of the first example:

```ruby
 # No direct equivalent to this in Sequel, but if you use the
 # tactical_eager_loading plugin, records will be preloaded
 # automatically as a set when needed.  Assuming you use
[...]

PR #84, opened 2018-05-14
example(RecordLoader): Add support for a unique parameter

razertory:
This change will allow to load_many in the RecordLoader

### DIFF
--- a/examples/record_loader.rb
+++ b/examples/record_loader.rb
@@ -1,9 +1,12 @@
 class RecordLoader < GraphQL::Batch::Loader
-  def initialize(model, column: model.primary_key, where: nil)
+  EMPTY_ARRAY = [].freeze
+
+  def initialize(model, column: model.primary_key, where: nil, unique: true)
     @model = model
-    @column = column.to_s
-    @column_type = model.type_for_attribute(@column)
+    @column = column
+    @column_type = model.type_for_attribute(@column.to_s)
     @where = where
+    @unique = unique
   end
 
   def load(key)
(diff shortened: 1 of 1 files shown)


examples/record_loader.rb
```diff
@@ -1,18 +1,31 @@
...
   def perform(keys)
-    query(keys).each { |record| fulfill(record.public_send(@column), record) }
-    keys.each { |key| fulfill(key, nil) unless fulfilled?(key) }
+    if @unique
+      query(keys).each do |record|
+        fulfill(record.public_send(@column), record)
+      end
+      default = nil
+    else
+      query(keys).group_by(&@column).each do |key, result|
```
razertory: * `group by` may not be a good choice without unique index (see: https://stackoverflow.com/questions/34115174/error-related-to-only-full-group-by-when-executing-a-query-in-mysql)

examples/record_loader.rb
```diff
@@ -1,18 +1,31 @@
 class RecordLoader < GraphQL::Batch::Loader
-  def initialize(model, column: model.primary_key, where: nil)
+  EMPTY_ARRAY = [].freeze
+
+  def initialize(model, column: model.primary_key, where: nil, unique: true)
     @model = model
-    @column = column.to_s
-    @column_type = model.type_for_attribute(@column)
+    @column = column
+    @column_type = model.type_for_attribute(@column.to_s)
```
razertory: * a `sequel model` has no `type_for_attribute` method

examples/record_loader.rb
dylanahsmith: This should be doing Array#group_by.  Perhaps `.to_a` should be used at the end of `query` to make this clear

examples/record_loader.rb
dylanahsmith: Right, that is why in https://github.com/Shopify/graphql-batch/issues/82#issuecomment-388565399 I said

> Likely the column type casting would still need to be removed to work with Sequel.

ayrton: What would that look like with Sequel?

PR #94, opened 2019-02-04
Remove deprecated execution strategies

swalkinshaw:
This removes the long since deprecated execution strategies method.
graphql gem minimum version has been bumped to >= 1.3.

Supporting these strategies is becoming harder with the class-based API
introduced in 1.8 and the upcoming interpreter changes in 1.9.


swalkinshaw:
@dylanahsmith CI was failing on 2.3 due to a bundler dependency version issue. I've removed `bundler` from the gemspec and also took this chance to test on newer Ruby versions.

PR #97, opened 2019-02-05
test: Remove no longer needed schema_execute helper method

dylanahsmith:
GraphQL::ExecutionStrategyTest used to override schema_execute to share the GraphQL::GraphQLTest tests before https://github.com/Shopify/graphql-batch/pull/94 removed the old execution strategies.  So now we can call execute on the schema directly.


dylanahsmith:
I guess we might end up wanting to bring this back to test the interpreter along with the current default execution strategy.

Issue #93, opened 2019-01-29
graphql-ruby 1.9 interpreter support


swalkinshaw:
1.9 offers a new optional interpreter and AST-based analysis which has breaking changes. Field instrumentation won't be supported and graphql-batch uses it for mutations: https://github.com/Shopify/graphql-batch/blob/d1244b7f0c649e23ea0d30c3a1d69edc28405bb2/lib/graphql/batch/setup.rb#L12-L25

Ideally we'd support both and conditionally check the gem version. Here's one idea on how to support 1.9 borrowed from @rmosolgo :

```ruby
def resolve_with_support(**kwargs)
  GraphQL::Batch::Executor.current.clear
  result = super
  context.schema.sync_lazy(result)
ensure
  GraphQL::Batch::Executor.current.clear
end
```

We'll have to figure out a good way of hooking into `resolve_with_support`. I'm not sure it's possible without some manual intervention (including a module for example).

rmosolgo:
> including a module for example

That's what came to mind for me too, the library requires the same thing for subscription root in the new runtime: 

https://github.com/rmosolgo/graphql-ruby/blob/master/spec/graphql/subscriptions_spec.rb#L112-L114

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/subscriptions/subscription_root.rb

dylanahsmith:
> I'm not sure it's possible without some manual intervention (including a module for example).

Possibly the module could be included automatically, although perhaps that is too invasive.  At least the field instrumentation is only needed for the mutation root object type.

> the library requires the same thing for subscription root in the new runtime

Currently that works through overriding `field`.  That doesn't seem like a general replacement for field instrumentation, since it would miss interface fields.  However, in this case it would only assume that the mutation root doesn't implement any interfaces, which seems likely to be true.

swalkinshaw:
That subscription root example is a little different I think since it works via field extensions. The example I had above just overrides `resolve_with_support` so it would need to be included in a base mutation (and the gem recommends always inheriting from your own base classes).

```ruby
class BaseMutation < GraphQL::Schema::Mutation
  include GraphQL::Batch::Mutation
end

class ProductCreate < BaseMutation
  # etc
end
```

dylanahsmith:
That is just a less generic replacement for field instrumentation which relies on an external resolver to be used.  As https://graphql-ruby.org/mutations/mutation_root.html says

> This object is defined like any other GraphQL object

[...]

PR #99, opened 2019-02-11
Try using a field extension for 1.9 interpreter

rmosolgo:
~~This doesn't quite work yet, and I think it's because of the changes in https://github.com/rmosolgo/graphql-ruby/pull/1984. (Previously, there were cases when execution would proceed eagerly _down_ the tree, but in that PR, it was changed to _always_ resolve each level of depth before proceeding to further depths.)~~ 

~~I think this means that the interpreter has to _hardcode_ handling for mutation root fields, which isn't too surprising. I'll take a look.~~ Fixed by https://github.com/rmosolgo/graphql-ruby/pull/2097

This adds a CI build for 1.9.0.pre4 and and implementation that uses a FieldExtension. I think it will work pretty well, the only problem will be if people do: 

```ruby
# DONT DO THIS, IT WOULD BE BAD 👻
use GraphQL::Batch
mutation(Types::Mutation)
```

It will still work _for now_ because plugins are applied _after_ other configurations. But in the future, there will be no "build" step, and instead, the schema will be constructed just as configured in the source, so it might _start_ breaking. 

We _could_ override `self.mutation(t)` to raise an error, to avoid that configuration error. Would you like to? 

(Also, I realize you may prefer a different path, depending on the conversation over at #93, but I thought I'd code one up just to see how it played out.)

### DIFF
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -46,6 +46,3 @@ require_relative "batch/loader"
 require_relative "batch/executor"
 require_relative "batch/setup"
 require_relative "batch/setup_multiplex"
-if defined?(GraphQL::Schema::FieldExtension)
-  require_relative "batch/mutation_field_extension"
-end
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -17,7 +17,16 @@ module GraphQL
 
     def self.use(schema_defn, executor_class: GraphQL::Batch::Executor)
       schema = schema_defn.target
-      if GraphQL::VERSION >= "1.6.0"
+      if schema.respond_to?(:interpreter?) && schema.interpreter?
+        if schema.mutation
+          schema.mutation.fields.each do |name, f|
+            field = f.metadata[:type_class]
+            field.extension(GraphQL::Batch::MutationFieldExtension)
+          end
+        end
+        instrumentation = GraphQL::Batch::SetupMultiplex.new(schema, executor_class: executor_class)
+        schema_defn.instrument(:multiplex, instrumentation)
+      elsif GraphQL::VERSION >= "1.6.0"
         instrumentation = GraphQL::Batch::SetupMultiplex.new(schema, executor_class: executor_class)
         schema_defn.instrument(:multiplex, instrumentation)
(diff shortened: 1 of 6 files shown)


swalkinshaw:
I'm in favour of is even if it ends up being a shorter term solution 👍It would be good to have a version that supports 1.9 before we consider a larger re-org or unification?

rmosolgo:
Same here, I think it'd be great to release a graphql-batch version that will support 1.9.0 so that people can update OK. Anything else I can do on this one, or does it look OK?

lib/graphql/batch.rb
```diff
@@ -17,7 +17,16 @@ def self.batch(executor_class: GraphQL::Batch::Executor)
 
     def self.use(schema_defn, executor_class: GraphQL::Batch::Executor)
       schema = schema_defn.target
-      if GraphQL::VERSION >= "1.6.0"
+      if schema.respond_to?(:interpreter?) && schema.interpreter?
```
dylanahsmith: I actually find comparing versions to be easier to maintain.  That way it is easy to remove dead code when we bump the minimum supported graphql gem version.  It also makes it easier to debug breaking changes like if `schema.respond_to?(:interpreter?)` stopped returning `true`, since it wouldn't fallback to code for older versions.

```suggestion
      if Gem::Version.new(GraphQL::VERSION) >= Gem::Version.new('1.9.0.pre3')
        require_relative "batch/mutation_field_extension"
```

Feature detection makes more sense when trying to support distinct implementations of something, but here we depend on a specific gem.

By moving the require for `MutationFieldExtension` into here, we can both avoid the extra version check as well as avoid loading that code when the interpreter isn't being used.

dylanahsmith: Other than that, LGTM

swalkinshaw: Someone could be using 1.9 though without the interpreter so this is a little different behaviour. However, I think it's safe to do anyway since field extensions work regardless of the interpreter. 👍 to the require in here as well

rmosolgo:
:+1: Thanks for taking a look! I applied that suggestion and cleaned up the now-needless `require` at the end of the file.

PR #103, opened 2019-02-22
Fix compatibility with non-class based mutation root object

### LINKED ISSUE #101: NoMethodError: undefined method `extension' for nil:NilClass when resolving field
After upgrading to the latest versions of graphql and graphql-batch, the resolvers fail to work when they resolve to a lazy value.

```
› rg graphql Gemfile.lock
154:    graphql (1.9.1)
155:    graphql-batch (0.4.0)
```

Error:

```
NoMethodError: undefined method `extension' for nil:NilClass
graphql-batch-0.4.0/lib/graphql/batch.rb:25:in `block in use'
graphql-batch-0.4.0/lib/graphql/batch.rb:23:in `each'
graphql-batch-0.4.0/lib/graphql/batch.rb:23:in `use'
graphql-1.9.1/lib/graphql/define/defined_object_proxy.rb:28:in `use'
graphql-1.9.1/lib/graphql/schema.rb:781:in `block (2 levels) in to_graphql'
graphql-1.9.1/lib/graphql/schema.rb:777:in `each'
graphql-1.9.1/lib/graphql/schema.rb:777:in `block in to_graphql'
graphql-1.9.1/lib/graphql/define/instance_definable.rb:161:in `instance_eval'
graphql-1.9.1/lib/graphql/define/instance_definable.rb:161:in `ensure_defined'
graphql-1.9.1/lib/graphql/schema.rb:265:in `define'
graphql-1.9.1/lib/graphql/define/instance_definable.rb:130:in `redefine'
graphql-1.9.1/lib/graphql/schema.rb:776:in `to_graphql'
graphql-1.9.1/lib/graphql/schema/member/accepts_definition.rb:120:in `to_graphql'
graphql-1.9.1/lib/graphql/schema.rb:718:in `graphql_definition'
graphql-1.9.1/lib/graphql/schema/field.rb:627:in `with_extensions'
graphql-1.9.1/lib/graphql/schema/field.rb:485:in `resolve'
```

Cause seems to be #99, and this code:
https://github.com/Shopify/graphql-batch/blob/77945d4857f51f05dd36c25d1c599c0cd613a032/lib/graphql/batch.rb#L20-L25

The `metadata` is empty (`{}`) on the field, so it cannot load a value to call `#extension` on. Making the `if` not match, so that part is skipped, makes everything work again.

The type that caused this error is defined using the new class-based syntax, but not all types in the schema is using that syntax. I don't know if it's relevant or not.

This happens both in my test setup (that uses some fixtures and stubs) and with real queries against the schema, so I don't think it's caused by any weird setup on my end.

We do not use the Interpreter and the schema is set up this way:

```ruby
module MyApiName
  class Schema < GraphQL::Schema
    query QueryRoot
    mutation MutationRoot

    use GraphQL::Batch
    use GraphQL::Backtrace
  end
end
```

Moving `use` on top of the other types (which is mentioned as a bad thing in #99) does not change anything either.

Is there anything else I can do here to help debug this? Is there a workaround I can use?

dylanahsmith:
I think this fixes #101, @Mange can you confirm that?
cc @rmosolgo 

## Problem

We assumed that we could always use field extensions with version 1.9 of the graphql gem but that doesn't work if the fields are defined on a mutation root object type that doesn't use the new class-based API.

## Solution

Make the use of field extensions conditional upon having a `schema.mutation.metadata[:type_class]`.  I also changed the test schema to use a non-class based mutation root when not testing with the interpreter.

I also noticed that the README documentation was saying that `use GraphQL::Batch` needed to follow `use GraphQL::Execution::Interpreter`, but it is actually the `mutation` line that it needs to follow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,9 @@ class RecordLoader < GraphQL::Batch::Loader
 end
 ```
 
-Use `GraphQL::Batch` as a plugin in your schema (for graphql >= `1.5.0`).
+Use `GraphQL::Batch` as a plugin in your schema _after_ specifying the mutation
+so that `GraphQL::Batch` can extend the mutation fields to clear the cache after
+they are resolved (for graphql >= `1.5.0`).
 
 ```ruby
 class MySchema < GraphQL::Schema
@@ -67,15 +69,6 @@ MySchema = GraphQL::Schema.define do
 
   GraphQL::Batch.use(self)
 end
-```
-
-##### With `1.9.0`'s `Interpreter` runtime
-
-Add `GraphQL::Batch` _after_ the interpreter, so that `GraphQL::Batch` can detect the interpreter and attach the right integrations:
-
-```ruby
-use GraphQL::Execution::Interpreter
-use GraphQL::Batch
 ```
 
 #### Field Usage
--- a/lib/graphql/batch.rb
+++ b/lib/graphql/batch.rb
@@ -17,20 +17,21 @@ module GraphQL
 
     def self.use(schema_defn, executor_class: GraphQL::Batch::Executor)
       schema = schema_defn.target
-      if Gem::Version.new(GraphQL::VERSION) >= Gem::Version.new('1.9.0.pre3')
-        require_relative "batch/mutation_field_extension"
+      if GraphQL::VERSION >= "1.6.0"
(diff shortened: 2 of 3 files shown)


Mange:
I'll try this later this week. I've some time allotted there to try the upgrade once again.

Thank you!

PR #112, opened 2019-08-04
Fix typo in README.md for Resolver Arguments

ksylvest:
Using `id` from keyword argument in resolver instead of undefined `args[:id]`.


swalkinshaw:
Good catch, thanks

PR #113, opened 2019-10-05
Document an HTTP Loader Example

ksylvest:
A sample implementation of a threaded HTTP loader using a connection pool. I think it might be be useful to document some of the things loaders might be used for.

### DIFF
--- a/examples/http_loader.rb
+++ b/examples/http_loader.rb
@@ -2,6 +2,7 @@
 #
 # 1. https://github.com/httprb/http
 # 2. https://github.com/mperham/connection_pool
+# 3. https://github.com/ruby-concurrency/concurrent-ruby
 #
 # Setup:
 #
@@ -43,13 +44,13 @@ module Loaders
     end
 
     def perform(operations)
-      threads = operations.map do |operation|
-        Thread.new do
+      futures = operations.map do |operation|
+        Concurrent::Promises.future do
           pool.with { |connection| operation.call(connection) }
         end
       end
       operations.each_with_index.each do |operation, index|
-        fulfill(operation, threads[index].value)
+        fulfill(operation, futures[index].value)
       end
     end
 
--- a/examples/http_loader.rb
+++ b/examples/http_loader.rb
@@ -1,0 +1,64 @@
+# A sample HTTP loader using:
+#
+# 1. https://github.com/httprb/http
+# 2. https://github.com/mperham/connection_pool
+#
+# Setup:
+#
+#   field :weather, String, null: true do
+#     argument :lat, Float, required: true
+#     argument :lng, Float, required: true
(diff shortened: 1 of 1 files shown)


razertory:
Appreciated for that. It would be better if documented the README file


| |
陈立春
|
|
邮箱：razertory@163.com
|

签名由 网易邮箱大师 定制

On 10/05/2019 13:10, Kevin Sylvestre wrote:

A sample implementation of a threaded HTTP loader using a connection pool. I think it might be be useful to document some of the things loaders might be used for.

You can view, comment on, or merge this pull request online at:

  https://github.com/Shopify/graphql-batch/pull/113

Commit Summary
Document an HTTP loader example
File Changes
Aexamples/http_loader.rb (64)
Patch Links:
https://github.com/Shopify/graphql-batch/pull/113.patch
https://github.com/Shopify/graphql-batch/pull/113.diff

examples/http_loader.rb
```diff
@@ -0,0 +1,64 @@
...
+  class HTTPLoader < GraphQL::Batch::Loader
+    def initialize(host:, size: 4, timeout: 4)
+      @host = host
+      @size = size
+      @timeout = timeout
+    end
+
+    def perform(operations)
+      threads = operations.map do |operation|
+        Thread.new do
```
chrisbutcher: This is great, thank you!

Just a quick question: it looks like `ConnectionPool` will correctly keep some number of initialized HTTP clients ready to perform requests. But is there a problem here if the number of `operations` is arbitrarily high, and hence we could spawn an arbitrary number of threads?

I wonder if https://github.com/ruby-concurrency/concurrent-ruby/blob/master/docs-source/thread_pools.md is also useful here?

ksylvest: @chrisbutcher good callout - I think it'd be fairly easy to exhaust the list (if some sort of complexity rules aren't applied):

https://graphql-ruby.org/queries/complexity_and_depth#prevent-complex-queries

I'll make some alterations tonight.

ksylvest: @chrisbutcher looking through the docs for thread pool within concurrent-ruby it looks like it'd require using a mutex to guard a critical section for saving results into a collection. 

Instead of taking that approach - any objection to using the higher-order promise concurrency paradigm?: 

https://github.com/Shopify/graphql-batch/pull/113/commits/23f8740556ff2189e4f222c4be02c0e0e81a9de0

chrisbutcher: Looks great to me. Thanks Kevin!

ksylvest:
Hi @chrisbutcher, happy Friday! Wanted to check if you'd have any objections to merging / anything else needed to close out this PR? Thanks!

chrisbutcher:
Looks great to me, thanks @ksylvest !

PR #120, opened 2020-02-01
Update CI to test both graphql 1.9 and 1.10

ksylvest:
Fix CI error with add_type when testing without TESTING_INTERPRETER:

    gems/graphql-1.10.2/lib/graphql/schema.rb:1767:in `add_type':
    Can't add legacy type: Mutation (GraphQL::ObjectType) (ArgumentError)

This new failure appears to be related to graphql-1.10.2 vs graphql-1.10.1. This technically deletes some coverage from this PR, but the history around the changes wasn't clear:

1. a6e58f626764aedba9e2b9858c91a64e454486dc (introduced TESTING_INTERPRETER to test using `GraphQL::Execution::Interpreter / GraphQL::Analysis::AST` 
2. 75bfef9b98faa15b59ee5df7f6ba1253083d7ee2 (fixed bug with non-class mutations)

Is it still standard to use a non-class mutation root / needed to test? If not can delete, otherwise may need to add support with 1.10.2.


swalkinshaw:
> Is it still standard to use a non-class mutation root / needed to test? If not can delete, otherwise may need to add support with 1.10.2.

Since this gem still supports the old legacy mutations we should ideally keep this test coverage and instead fix it. I think the first step is updating the Gemfile to use `1.10.2` instead of `1.10.0.pre2` and make sure everything is actually working.

ksylvest:
Trace for this error is:

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema.rb#L1767

It looks like this feature is **not** supported in the `graphql-ruby`  gem (can see the problem when bumping from v1.9.19 to v1.10.0 of GraphQL). 

Here's a minimal reproducing of the error:

```ruby
require 'graqphl'

MutationType = GraphQL::ObjectType.define do
  name "Mutation"
end

class Schema < GraphQL::Schema
  mutation MutationType
end
```

https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#deprecations-1

Given that graphql-ruby is not supporting does it still make sense to try and maintain this spec? Could version lock a release via this to indicate also:

```ruby
spec.add_runtime_dependency "graphql", ">= 1.10", "< 2"`
```

swalkinshaw:
Ah... the issue is that we're always getting 1.10 now that it's a non-pre release. We'll need to fix CI (and the conditional tests/Gemfile) to test two different versions: <1.10 (really 1.9) and then >= 1.10.

I think it's worthwhile to still have the current version (and the next release) of graphql-batch support both of those transparently.

ksylvest:
@swalkinshaw adjusted to test versions + renamed testing env. 

It sounds like you'd be open to merging a PR that drops support once the next version of graphql-batch goes live? If so - keeping this PR handy: https://github.com/ksylvest/graphql-batch/pull/1

.travis.yml
```diff
@@ -3,5 +3,9 @@ rvm:
   - 2.3
   - 2.6
 env:
-  - TESTING_INTERPRETER=true
-  - TESTING_INTERPRETER=false
+  -
+    GRAPHQL_VERSION: 1.9.19
+    TESTING_LEGACY_MUTATION: true
```
swalkinshaw: Just a naming nitpick but I'd probably generalize this a bit: `TESTING_LEGACY_DEFINITION_LAYER`

ksylvest: Amended.

swalkinshaw:
Thank you 🎉

PR #121, opened 2020-02-03
Drop support for define mutation tests

ksylvest:
The GraphQL ruby gem dropped support for define mutations between version v1.9.19 and v1.10.0. This bumps the required version to the latest and removes the legacy test. 

Note: initial proposed in https://github.com/Shopify/graphql-batch/pull/120

Per discussion - blocked until release is live with fix?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: Tests
 
-on:
-  push
+on: push
 
 jobs:
   test:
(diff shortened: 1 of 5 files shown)


swalkinshaw:
It's not just in tests, we have backwards compat code as well: https://github.com/Shopify/graphql-batch/blob/fe002252ce61822af91f47ceb5851f0d61a337f2/lib/graphql/batch.rb#L26-L32

I think we still want to wait a bit even with a new version. Because any other new features/fixes in the future wouldn't apply to anyone who needs 1.9 support (unless we backported them to other versions which be annoying).

So I'm good with doing this... but only after some time passes. Like 6 months for example.

ksylvest:
@swalkinshaw I wanted to followup on this and see if you think enough time has passed for users of graphql-batch to have upgraded? I've adjusted this PR (noticed the CI swapped to GH, and didn't seem to run). I'd be happy to make any changes needed, or close if your thoughts on timing have changed.

swalkinshaw:
Thanks for the followup! I think we can safely ship this now. graphql-batch is pretty stable and feature complete anyway/

Did CI not run on this PR at all? It might need to be re-created 🤔

ksylvest:
@swalkinshaw hmm - tried closing / re-opening https://github.com/Shopify/graphql-batch/pull/136. I'm not familiar with GH actions - but didn't see it appear.

PR #127, opened 2020-04-21
Fix ruby 2.7 keyword arguments warning

### LINKED ISSUE #126: Ruby 2.7 keyword arguments warning
I'm getting a warning running my project that uses latest version of graphql-batch:
```
/usr/local/bundle/gems/graphql-batch-0.4.2/lib/graphql/batch/loader.rb:14: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/app/lib/my_loader.rb:7: warning: The called method `initialize' is defined here
```

`MyLoader` initializer accepts kwargs but `Loader.for` method doesn't seem to pass them correctly to `new` method in here: https://github.com/Shopify/graphql-batch/blob/master/lib/graphql/batch/loader.rb#L14

dylanahsmith:
Fixes #126

## Problem

GraphQL::Batch::Loader was using `*group_args` for forwarding arguments which won't forward keyword arguments in the future which may be used by the loader (e.g. we do that in the RecordLoader example).

## Solution

Use `*group_args` for argument forwarding for older ruby versions, since passing keyword arguments an initializer that doesn't take them causes problems for old versions of ruby.

As an optimization, I used `...` syntax for argument forwarding for MRI ruby 2.7+ which avoids some object allocation for the splat parameters.  I had to eval the ruby code using the new syntax to avoid causing parsing problems for older versions of ruby.

I also changed the CI config to test ruby 2.7.

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -1,21 +1,20 @@
 module GraphQL::Batch
   class Loader
-    def self.for(*group_args)
-      loader_key = loader_key_for(*group_args)
-      executor = Executor.current
-
-      unless executor
-        raise GraphQL::Batch::NoExecutorError, 'Cannot create loader without'\
-          ' an Executor. Wrap the call to `for` with `GraphQL::Batch.batch`'\
-          ' or use `GraphQL::Batch::Setup` as a query instrumenter if'\
-          ' using with `graphql-ruby`'
+    # Use new argument forwarding syntax if available as an optimization
+    if RUBY_ENGINE && Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7")
+      class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+        def self.for(...)
+          current_executor.loader(loader_key_for(...)) { new(...) }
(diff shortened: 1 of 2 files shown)


lib/graphql/batch/loader.rb
```diff
@@ -1,21 +1,20 @@
...
+    # Use new argument forwarding syntax if available as an optimization
+    if RUBY_ENGINE && Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("2.7")
+      class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+        def self.for(...)
+          current_executor.loader(loader_key_for(...)) { new(...) }
+        end
+      RUBY
+    else
+      def self.for(*group_args, **group_kwargs)
+        current_executor.loader(loader_key_for(*group_args)) { new(*group_args) }
```
swalkinshaw, line 12: `group_kwargs` isn't passed on here?

dylanahsmith: Oops, I meant to remove it from the parameters so that it got captured in `*group_args`.  Pushed a fix

PR #128, opened 2020-04-29
Adding examples for Active Storage Loader and Postgres Window Loader

leighhalliday:
I love when a project has a ton of examples that can be used and extended. Example: https://github.com/zeit/next.js/tree/canary/examples

I wanted to add 2 examples that solve somewhat challenging batch issues:

- Loading `has_many` relationships where you want the first N records. This can be accomplished using window functions in Postgres.
- Loading Active Storage attachments (can often produce 2N + 1 queries).

### DIFF
--- a/examples/active_storage_loader.rb
+++ b/examples/active_storage_loader.rb
@@ -1,0 +1,52 @@
+####
+# This is a loader for has_one_attached Active Storage attachments
+# To load a variant for an attachment, 2 queries are required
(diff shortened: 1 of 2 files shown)


swalkinshaw:
👋 hey @leighhalliday, lost track of this but I agree it's useful to have more examples. I'm not going to verify these, but they look good so I'll merge.

PR #133, opened 2021-01-12
Add support to Ruby 3.0

lucianosousa:


swalkinshaw:
@lucianosousa we'll need to redo with on GitHub actions now. Though 3.0 is currently failing due within the graphql gem. I think we need to use a newer version for 3.0 support

lucianosousa:
Thanks for the feedback @swalkinshaw.

PR #134, opened 2021-01-12
Switch CI to GitHub actions

swalkinshaw:
Travis is useless now so let's switch


swalkinshaw:
@lucianosousa want to wait for this to land before we add 3.0?

RobertWSaunders (approved):
Makes sense to me! 🔥

PR #135, opened 2021-01-12
Add Ruby 3.0 to CI

swalkinshaw:
@lucianosousa

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,8 +9,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.3, 2.6, 2.7]
-        graphql_version: ['1.9.19', '1.10.2']
+        ruby: [2.3, 2.6, 2.7, 3.0]
+        graphql_version: ['1.9.19', '1.11.6']
+        exclude:
+          - graphql_version: '1.9.19'
+            ruby: 3.0
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,8 +9,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.3, 2.6, 2.7]
-        graphql_version: ['1.9.19', '1.10.2']
+        ruby: [2.3, 2.6, 2.7, 3.0]
```
eapache, line 12: If we're testing 2.3 and 3.0 now we probably don't need _both_ 2.6 and 2.7 here

swalkinshaw: 🤷‍♂️  sure let's drop 2.6

lucianosousa:
I'm doing some tests with Ruby 3.0 rails Rails master [here](https://github.com/lucianosousa/graphql-ruby/runs/1691360567?check_suite_focus=true) and it's not working...

Are we doing some tests using rails here?
I think the issue I saw was related to the GraphQL DB connection classes

swalkinshaw:
This gem doesn't have any integration tests using Rails, but from looking at those failures, it's not actually Rails/DB related; just the Relay connections in the graphql gem. So while graphql-batch may technically work with Ruby 3.0, most consumers would probably run into trouble with the graphql gem itself (unless they don't use connections 🤷‍♂️ ).

PR #136, opened 2021-02-24
Drop support for define mutation tests

ksylvest:
The GraphQL ruby gem dropped support for define mutations between version v1.9.19 and v1.10.0. This bumps the required version to the latest and removes the legacy test. 

Note: initial proposed in https://github.com/Shopify/graphql-batch/pull/120

This PR is a re-build for this https://github.com/Shopify/graphql-batch/pull/121 to fix CI (hopefully).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,6 @@
 name: Tests
 
-on:
-  push
+on: push
 
 jobs:
   test:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,7 +1,6 @@
 name: Tests
 
-on:
-  push
+on: push
```
swalkinshaw, line 3: Oh I think we need `on: [push, pull_request]` since push on works on branches *on this repo* and yours is a fork obviously.

ksylvest: Eureka - that fixed! Thanks Scott. CI appears happy.

lib/graphql/batch.rb
```diff
@@ -15,31 +15,21 @@ def self.batch(executor_class: GraphQL::Batch::Executor)
...
+      schema.instrument(:multiplex, instrumentation)
+      if schema.mutation
+        if schema.mutation.is_a?(Class) || schema.mutation.metadata[:type_class]
+          require_relative "batch/mutation_field_extension"
+          schema.mutation.fields.each do |name, f|
+            field = f.respond_to?(:type_class) ? f.type_class : f.metadata[:type_class]
+            field.extension(GraphQL::Batch::MutationFieldExtension)
           end
+        else
+          schema.instrument(:field, instrumentation)
```
swalkinshaw, line 29: 🤔  there's actually a lot more we can get rid of now by dropping support for < 1.10 (and hence legacy non-class based types).

This conditional would go away entirely and the older `GraphQL::Batch::Setup` would be dropped and merged into `SetupMultiplex`: https://github.com/Shopify/graphql-batch/blob/9fa2dd255e7d0801271471566e3c80083f0adbea/lib/graphql/batch/setup.rb#L2

(`SetupMultiplex` is newer but references methods from `Setup`)

Do you want to take care of those as well in this PR?

PR #138, opened 2021-03-25
Switch to the rubocop-shopify gem

dylanahsmith:
Our rubocop config is still trying to inherit the shopify rubocop config from https://ruby-style-guide.shopify.dev/rubocop.yml which now gives a 404 error.

The recommended way of using this configuration has been to use the rubocop-shopify gem, as mentioned on https://ruby-style-guide.shopify.dev/#adoption-with-rubocop.  So I've switched to using that gem and fixed some rubocop offenses from inheriting a newer config.


.rubocop_todo.yml
```diff
@@ -56,6 +56,13 @@ Layout/SpaceInsideParens:
   Exclude:
     - 'test/loader_test.rb'
 
+# Offense count: 3
+Lint/MissingSuper:
+  Exclude:
+    - 'test/executor_test.rb'
+    - 'test/loader_test.rb'
+    - 'test/support/loaders.rb'
```
dylanahsmith, line 64: Although not calling `super` isn't recommended, we do need to keep supporting it for backwards compatibility, so I avoided updating these offenses.

swalkinshaw (approved):

PR #140, opened 2021-06-03
Fix errors in examples

ham0215:
When using the loader of examples, the following error occurred.

```
wrong number of arguments (given 2, expected 0)
/app/app/loader/association_loader.rb:10:in `initialize'
```

The problem is passing arguments to the `super`, so I fixed it so that no arguments are passed.


tgwizard:
Thanks!

Issue #145, opened 2021-10-29
fulfilled? always returns false even if it's fulfilled with promise


mtsmfm:
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'graphql', '1.12.17'
  gem 'graphql-batch', github: 'Shopify/graphql-batch'
end

class UserLoader < GraphQL::Batch::Loader
  def perform(ids)
    ids.select {|id| id != 'invalid' }.each do |id|
      fulfill(id, EchoLoader.load(id))
    end

    ids.each do |id|
      p [id, fulfilled?(id)]
      fulfill(id, nil) unless fulfilled?(id)
    end
  end
end

class EchoLoader < GraphQL::Batch::Loader
  def perform(keys)
    keys.each { |key| fulfill(key, key) }
  end
end

class UserType < GraphQL::Schema::Object
  field :id, ID, null: true
  field :name, String, null: true

  def id
    object
  end

  def name
    UserLoader.load(object)
  end
end

class QueryType < GraphQL::Schema::Object
  field :users, [UserType], null: false do
    argument :ids, [ID], required: true
  end

  def users(ids:)
    ids
  end
end

class AppSchema < GraphQL::Schema
  query QueryType
  use GraphQL::Batch
end

p AppSchema.execute(<<~Q).to_h
query {
  users(ids: ["alice", "bob"]) {
    id, name
  }
}
Q
```

Above code outputs

```
["alice", false]
["bob", false]
{"data"=>{"users"=>[{"id"=>"alice", "name"=>nil}, {"id"=>"bob", "name"=>nil}]}}
```

I think it should be

```
["alice", true]
["bob", true]
{"data"=>{"users"=>[{"id"=>"alice", "name"=>"alice"}, {"id"=>"bob", "name"=>"bob"}]}}
```

Is it intentional?

mtsmfm:
It seems it's fixed by 0.5.0

PR #146, opened 2021-11-17
Make Loader#fulfilled? return true if fulfilled with another promise

leonidkroka:
## Problem
`Loader#fulfilled?` returns `false` if it's fulfilled with promise

## Solution
Check if `Promise#source` was changed


leonidkroka:
> This is a follow-up to #143 right?
Yes, you are right. Also, the issue is well described [here](https://github.com/Shopify/graphql-batch/issues/145) 

> Is it possible to add a test for this behavior?
Added two fulfilling status checking tests

swalkinshaw:
Sorry for the delay. Thanks for this contribution @leonidkroka

PR #147, opened 2021-11-20
Fix AssociationLoader Rails 7 Deprecation Warning

keoghpe:
Currently running this example code with rails 7 yields the following warning:

```
DEPRECATION WARNING: `preload` is deprecated and will be removed in Rails 7.0. Call `Preloader.new(kwargs).call` instead.
```

Updating the preloader code to use kwargs and call silences the warning.

### DIFF
--- a/examples/association_loader.rb
+++ b/examples/association_loader.rb
@@ -36,7 +36,7 @@ class AssociationLoader < GraphQL::Batch::Loader
   end
 
   def preload_association(records)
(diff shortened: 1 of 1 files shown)


swalkinshaw:
I'm thinking this should wait until Rails 7.0.0 is actually released since `Preloader.new(kwargs).call` doesn't exist in any prior stable versions.

PR #150, opened 2022-02-09
Prep for graphql-ruby 2.0

rmosolgo:
I'm removing a bunch of deprecated code in graphql-ruby 2.0 and I want to make sure graphql-batch will be good-to-go on that version. 

I ran the tests with graphql 1.13.x. These two configs gave warnings (to remove them), so I removed them. Then I pointed this project's gemfile to use `rmosolgo/graphql-ruby` (the 2.0 branch is on master now), and everything worked fine :+1: So, I think the gem's source is 2.0-ready, and just the tests needed an update.

### DIFF
--- a/test/support/schema.rb
+++ b/test/support/schema.rb
@@ -167,7 +167,5 @@ class Schema < GraphQL::Schema
   query QueryType
   mutation MutationType
 
-  use GraphQL::Execution::Interpreter
-  use GraphQL::Analysis::AST
   use GraphQL::Batch
 end


swalkinshaw (approved):
Ah I left these in since I thought 1.10 might have required them still, but obviously not or the test for that version would have failed 👍

rmosolgo:
(Sorry, if I'd realized the whole test setup, I would've done some `if`s, etc instead of outright removal!)

swalkinshaw:
> so my best guess is that the 1.10 test is using the old runtime for that suite (and graphql-batch supports both fine)

I removed that though 🤔 🤔 🤔 in https://github.com/Shopify/graphql-batch/pull/148

swalkinshaw:
Ah thanks. I'll go forward with a release then 👍

PR #151, opened 2022-02-09
Allow graphql 2.0, add it to the build

rmosolgo:
I'm not sure how y'all actually want to use this version parameter, but I noticed it prevents GraphQL-Ruby 2.0. I added that new version to the build to double-check that it still worked, and while I was there, also bumped the top ruby version. Feel free to take any of this that you like, or let me know if there's anything I can change to make it more helpful!


swalkinshaw:
Makes sense. This was just a conservative guess initially I think.

PR #153, opened 2023-01-16
Improve Shapes Friendliness of Batch::Loader

casperisfine:
For optimal performance in Ruby 3.2, it is preferable to define instance variables in a consistent order.

Based on production data, it seems that Batch::Loader is one of the main offenders in our app:

```
Shape Edges Report
-----------------------------------
       169  @cache
       130  @queue
       127  @executor
       125  @sql_counter
       119  @loader_key
```

### DIFF
--- a/lib/graphql/batch/loader.rb
+++ b/lib/graphql/batch/loader.rb
@@ -44,6 +44,13 @@ module GraphQL::Batch
 
     attr_accessor :loader_key, :executor
 
+    def initialize
+      @loader_key = nil
+      @executor = nil
+      @queue = nil
+      @cache = nil
+    end
+
     def load(key)
       cache[cache_key(key)] ||= begin
         queue << key


swalkinshaw (approved):

casperisfine:
> Is this a quirk with Shapes and/or YJIT that might be ironed out in the future, or a performance optimization we should consider for our own apps and gems?

It's a performance optimization. Generating too many shapes prevent caching instance variable access and efficient YJIT compilation. So it's not that it makes things worse than they were on 3.1, just that they prevent it from performing better.

> but a small writeup or blog post would probably be of interest to a lot of people like me!

Right, I'm quite swamped right now, but that's definitely something we need to do. In the meantime Jemma's talk will likely satisfy your curiosity: https://www.youtube.com/watch?v=So-KvN3p-eE

trevorturk:
Excellent, thank you for the info!

PR #155, opened 2023-02-14
Add Ruby 3.2 to CI

dipth:


dipth:
It appears that I need to rerun the CLA workflow, but it doesn't appear that I have any way to do so...

swalkinshaw:
@dipth done. Looks like there's a legit failure on 3.2 due to the usage of refinements.

dipth:
I would personally lean towards bumping the `graphql` dependency, but it's your call 🙂

swalkinshaw:
Yeah I'm good with that. Would be interested in doing that?

dipth:
Closed in favor of:
https://github.com/Shopify/graphql-batch/pull/156

PR #156, opened 2023-02-14
Add support for Ruby 3.2

dipth:
These changes adds support for Ruby 3.2 by bumping the minimum required version of `graphql` from `1.10.0` to `1.12.18` which fixes a Ruby 3.2 incompatibility related to the use of include in refinements as fixed in this PR:
https://github.com/rmosolgo/graphql-ruby/pull/3674

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


swalkinshaw:
Excellent, thanks @dipth

dipth:
You're right. We're doing a number of dependency upgrades ahead of upgrading to Ruby 3.2, so I must have mixed this one together with another one that did have actual code changes :)

PR #164, opened 2024-01-15
Use trace_with to start and end batches

rmosolgo:
I'm hoping to consolidate GraphQL-Ruby's many instrumentation hooks in GraphQL-Ruby 3.0 (someday). I'm hoping to add a deprecation warning to `instrument(...)` in the next release (https://github.com/rmosolgo/graphql-ruby/pull/4771), `trace_with` is sufficient for those cases. 

This PR moves GraphQL-Batch's hooks to `trace_with`, what do you think of that?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.4, 2.7, '3.1', '3.2']
+        ruby: [2.7, 3.1, 3.3]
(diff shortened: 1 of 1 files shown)


rmosolgo:
It looks like GraphQL-Ruby doesn't support Ruby 2.4 anymore 😅

swalkinshaw:
Fixed in https://github.com/Shopify/graphql-batch/pull/165

rmosolgo:
Thanks, I merged master into this branch :+1:

mgruner:
@swalkinshaw thank you very much. Do you plan to release this in a new version? That would be extremely helpful.

PR #167, opened 2024-02-20
Make executor_class optional in trace class

swalkinshaw:
Otherwise `Schema.new_trace` will require `executor_class` as a keyword arg which is the wrong place to pass it in.

@rmosolgo should `execute_multiplex` check for the existence of `executor_class` or is this existing one enough? https://github.com/Shopify/graphql-batch/blob/8e88b28d10a894d3bbb221095424fc948e62bccd/lib/graphql/batch/executor.rb#L22-L24

### DIFF
--- a/lib/graphql/batch/setup_multiplex.rb
+++ b/lib/graphql/batch/setup_multiplex.rb
@@ -14,7 +14,7 @@ module GraphQL::Batch
     end
 
     module Trace
-      def initialize(executor_class:, **_rest)
+      def initialize(executor_class: nil, **_rest)
         @executor_class = executor_class
         super
       end


rmosolgo:
I _think_ this was in the right place. I ran into a bug here too, but I think it was a bug in GraphQL-Ruby (It was with `backtrace: true`), which I addressed in https://github.com/rmosolgo/graphql-ruby/pull/4849

Did you encounter an error here too?

rmosolgo:
Oh, OK -- different bug 😅 I think this will fix it: https://github.com/rmosolgo/graphql-ruby/pull/4850

(It _should_ just work -- the option given to the `default` mode should have been available to downstream, derived modes, too. The trace module _was_ mixed in, but the options hash wasn't properly merged.)

swalkinshaw:
looks good, thank you 🎉

PR #168, opened 2024-02-22
Priming the cache

tgwizard:
Sometimes data is loaded through some means, which will also later be loaded by an independent batch data loader. It can be convenient to then prime the cache of that data data loader with the already loaded data, to remove redundant calls.

The use-case I have can be summarized by this query:

```
query {
  shopsFollowed(first: 10) {
    nodes {
      id
      followedByMe
    }
  }
}
```

The `shopsFollowed` resolver performs a query to load 10 shops that the current user follows. This is returned as the `Shop` type, which is used extensively in the API. The `Shop.followedByMe` field returns a boolean of whether the current user follows the shop. It's backed by a custom batch data loader.

For the specific `shopsFollowed` use-case, we know that all the shops returned are followed by the current user, so there's no need for the custom batch data loader to load this again.

With `prime` we can have the `shopsFollowed` resolver pre-populate this information on the loader.

This is available in other graphql batch data loader libraries, see e.g.

 - https://github.com/graphql/dataloader?tab=readme-ov-file#primekey-value
 - https://github.com/graph-gophers/dataloader/blob/ab736ad423a90b8560fb932a3c523693b52c8f35/dataloader.go#L369

### DIFF
--- a/test/loader_test.rb
+++ b/test/loader_test.rb
@@ -215,4 +215,40 @@ class GraphQL::Batch::LoaderTest < Minitest::Test
     promise = ExplodingLoader.load([1]).then(nil, ->(err) { error_message = err.message } ).sync
     assert_equal 'perform failed', error_message
   end
+
+  def test_prime
+    loader = EchoLoader.for
+    loader.prime(:a, :prepared_value)
+
+    assert_equal :prepared_value, loader.load(:a).sync
+    assert_equal [:prepared_value, :b, :c], loader.load_many([:a, :b, :c]).sync
+  end
+
+  def test_will_not_call_perform_if_fully_primed
+    loader = ExplodingLoader.for
+    loader.prime(:a, 1)
+    loader.prime(:b, 2)
+
+    assert_equal [1, 2], loader.load_many([:a, :b]).sync
+  end
+
+  def test_priming_a_key_already_in_queue_does_nothing
+    loader = EchoLoader.for
+
+    promise = loader.load(:a)
+
+    loader.prime(:a, :not_a)
+
+    assert_equal :a, promise.sync
+  end
+
+  def test_prime_will_not_replace_already_cached_value
+    loader = EchoLoader.for
+
+    assert_equal :a, loader.load(:a).sync
+
+    loader.prime(:a, :not_a)
+
+    assert_equal :a, loader.load(:a).sync
+  end
 end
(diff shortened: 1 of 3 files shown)


test/loader_test.rb
```diff
@@ -215,4 +215,30 @@ def test_loader_with_failing_perform
...
+
+  def test_prime_will_not_replace_already_cached_value
+    loader = EchoLoader.for
+
+    assert_equal :a, loader.load(:a).sync
+
+    loader.prime(:a, :not_a)
+
+    assert_equal :a, loader.load(:a).sync
+  end
```
michaelherold, line 243: **comment:** This seems like it potentially is a footgun for surprising behavior. Particularly with ActiveRecord's lazy loading of associations, what if you try to prime with a model that has preloaded an association that you need (N.B. if you're using GraphQL::Batch, you probably shouldn't be loading associations like this, but stranger things have happened), but it has already been fulfilled with a version that does not have that association preloaded?

**question:** Should we emit a warning or something by default when attempting to prime an already-fulfilled promise? One could then opt in to allow either overriding or ignoring the warning.

This could very well be overkill but it's the first thing that came to mind when I saw [the implementation](https://github.com/Shopify/graphql-batch/pull/168/files#diff-1c95774a012e8c42cb5ac276b32dce24c7cb3997b6ef9118794d75af24c75256R67).

michaelherold (commented):
I like how simple this is. I wonder if the simplicity could potentially hide performance bugs. What do you think?

test/loader_test.rb
tgwizard: Yes, this is indeed not meant to be used to populate the loader with values that are in a certain state, for some kind of perf reasons. The values you put in should behave the same as the values the loader would have loaded on its own.

I'm not sure there's a good and simple way to protect against this, except in documenting its intended use.

Given this, I don't think we should emit a warning when priming a value for a key that already exists in the loader cache. That is meant to be a no-op - loading or priming **should** result in equivalent results - so it shouldn't seem to the developer that they are doing something wrong. Given the resolution of GraphQL it can be very hard to reason or know/ensure that you'd be priming keys that won't exist.

The behaviour here (priming inserts a key in the cache only if the key didn't exist) is also the behaviour of the other data loader libraries I found.

What do you think?

swalkinshaw (approved):

PR #171, opened 2024-04-03
Centralize ruby version

george-ma:


swalkinshaw:
@george-ma this doesn't actually add a `.ruby-version` file. Should I add it manually?

graphql-batch.gemspec
```diff
@@ -10,6 +10,8 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/Shopify/graphql-batch"
   spec.license       = "MIT"
 
+  spec.required_ruby_version = ">= 2.4"
```
swalkinshaw, line 13: @george-ma we have tests using 2.4 https://github.com/Shopify/graphql-batch/blob/8afcdbbfbc6027d75d3fd496cffe97c1ce822e30/.github/workflows/ci.yml#L16-L19

So I updated this to 2.4 instead of 2.7, but now Rubocop requires 2.7. Was there a reason why you had to bump rubocop?

PR #207, opened 2026-03-18
Fix loader for kwargs mutation ruby4

AlfonsoUceda:
Hello!

I'm creating this PR because I've found a problem using this gem with Ruby 4 and a specific way of using loaders. The problem seems with the use of `...` operator forwarding in ruby 4 and the loader class.

On Ruby 4.0, when kwargs are destructured by a caller method and re-passed to Loader.for(...), the ... forwarding shares the internal kwargs hash across all expansion sites. When new(...) destructures them in initialize, it empties the same hash already stored in the executor's loader key, breaking loader reuse.

This only manifests when .for() is called through an intermediary that captures kwargs as named parameters and re-passes the, a common pattern in subclasses:

```ruby
  def self.load(model_class, id, key: :id, preload: true, **conditions)
    self.for(model_class, key: key, preload: preload, **conditions).load(id)
  end
```
`Direct .for(key: 'value')` calls are unaffected, which is why the bug is subtle.

1. loader_key_for(...) captures kwargs into the key: [self, {key: :id, ...}, [model]]
2. new(...) in the block shares the same kwargs hash - initialize destructures it, emptying the hash
3. The stored key becomes [self, {}, [model]]
4. Next lookup with {key: :id, ...} doesn't match {} - a new loader is created.

I've made thanks to Claude that helped me to identify this problem in our codebase and pointed out to this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby: [2.7, 3.0, 3.1, 3.2, 3.3]
+        ruby: [2.7, 3.0, 3.1, 3.2, 3.3, 3.4, 4.0]
         graphql_version: ['~> 1.13', '~> 2.0']
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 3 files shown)


AlfonsoUceda:
I have signed the CLA!

swalkinshaw:
@AlfonsoUceda thanks for the issue + fix. I've opened an alternative in https://github.com/Shopify/graphql-batch/pull/216 because I realized we can clean up the whole version conditional support if we drop 2.7 support

AlfonsoUceda:
thanks @swalkinshaw ;)

PR #215, opened 2026-04-23
Support GraphQL-Ruby's upcoming execute_next APIs

rmosolgo:
I've been working on a new breadth-first execution flow for GraphQL-Ruby based on Shopify's in-house work in that area. I'm just about ready to release it but I realized that GraphQL-Batch's field extension would need an update. 

The new execution flow passes `objects: ...` instead of `object:` to this method. I made the method take both arguments, but require neither one, so that the same method would work on either _new_ or _old_ execution. 

I also added a CI build to use Ruby 4.0 and `execute_next` by default.

I used https://github.com/rmosolgo/graphql_migrate_execution to migrate `test/support/schema.rb`.

TODO: 

- [x] Release a new gem version with fixes (already on master) for errors raised during promises

### DIFF
--- a/lib/graphql/batch/mutation_field_extension.rb
+++ b/lib/graphql/batch/mutation_field_extension.rb
@@ -3,10 +3,10 @@ module GraphQL::Batch
     def resolve(object: nil, objects: nil, arguments:, **_rest)
       GraphQL::Batch::Executor.current.clear
       begin
-        if object
-          ::Promise.sync(yield(object, arguments))
-        else
+        if !objects.nil?
           ::Promise.sync(yield(objects, arguments))
+        else
+          ::Promise.sync(yield(object, arguments))
         end
       ensure
         GraphQL::Batch::Executor.current.clear
--- a/lib/graphql/batch/mutation_field_extension.rb
+++ b/lib/graphql/batch/mutation_field_extension.rb
@@ -1,9 +1,13 @@
 module GraphQL::Batch
   class MutationFieldExtension < GraphQL::Schema::FieldExtension
-    def resolve(object:, arguments:, **_rest)
+    def resolve(object: nil, objects: nil, arguments:, **_rest)
       GraphQL::Batch::Executor.current.clear
       begin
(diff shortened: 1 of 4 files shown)


rmosolgo:
I expect this to pass CI on GraphQL-Ruby 2.5.24

swalkinshaw:
@rmosolgo great timing, I just added 4.0 to CI so you'll need to rebase

rmosolgo:
Got it -- rebased!

lib/graphql/batch/mutation_field_extension.rb
```diff
@@ -1,9 +1,13 @@
...
   class MutationFieldExtension < GraphQL::Schema::FieldExtension
-    def resolve(object:, arguments:, **_rest)
+    def resolve(object: nil, objects: nil, arguments:, **_rest)
       GraphQL::Batch::Executor.current.clear
       begin
-        ::Promise.sync(yield(object, arguments))
+        if object
+          ::Promise.sync(yield(object, arguments))
+        else
+          ::Promise.sync(yield(objects, arguments))
```
swalkinshaw: I guess this doesn't make a functional difference, but since `objects` is the new edge case, I'd think we should check that as the condition?

rmosolgo: :+1: Sounds good. Actually I got burned in graphql-ruby for forgetting that `false` can be valid runtime value -- this code might have misbehaved in that case. 

I updated it to check `!objects.nil?` in 7576a75, let me know what you think of that.

swalkinshaw: Makes sense 👍

swalkinshaw:
@rmosolgo the workflow fails on `main`: https://github.com/Shopify/graphql-batch/actions/runs/24898092897

rmosolgo:
🤦  Oops, hopefully fixing the typo will get us green again: https://github.com/Shopify/graphql-batch/pull/217

